PR #2, opened 2026-07-26
Add versioned causal observation artifacts

FlorianPfaff:
## What changed

- add a versioned `ObservationArtifact` contract for downstream Bayesian consumers
- record explicit metric versus gauge-relative status, gauge state, covariance units, source-window identities, correlation groups, and per-frame source limits
- add fail-closed validation for shapes, finiteness, contributor consistency, covariance symmetry and positive semidefiniteness, provenance, and causal boundaries
- add hash-locked JSON/NPZ serialization without pickle
- add `prob4d-observation export` and `prob4d-observation validate`
- make causal export discard future-dependent MotionCrafter windows and rerun gauge estimation and fusion rather than slicing an all-frame result
- document the artifact schema and causal semantics
- add focused regression tests for round trips, future-dependency rejection, gauge/metric consistency, indefinite covariance rejection, and payload tampering
- make the repository root explicit on pytest's import path so existing tests can import utilities from `scripts/`

## Why

Prob4D currently exports dense fused arrays, but Bayesian-PhysTwin and Causal4D need a narrow, auditable observation boundary. In particular, a downstream consumer must be able to distinguish metric observations from unresolved monocular gauge coordinates and verify that no source frame after a causal prefix influenced the product.

## Impact

Existing prediction manifests and fusion commands remain unchanged. The new command is additive. Standard MotionCrafter exports are deliberately labelled `gauge_relative` with covariance in `gauge_unit^2`; they cannot be silently treated as metric observations.

## Validation

- full GitHub Actions workflow passed
- Ruff passed on `src` and `tests`
- complete pytest suite passed
- focused observation-artifact test suite contains five regression tests
- Python byte-compilation passed in the isolated implementation check

The initial PR run exposed a pre-existing test-collection issue: two tests imported repository utilities from `scripts/`, but the repository root was not explicit on pytest's import path. The final diff fixes that in `pyproject.toml`; the temporary JUnit diagnostic workflow change was removed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,10 +15,4 @@ jobs:
           cache: pip
       - run: python -m pip install -e ".[dev]"
       - run: ruff check src tests
-      - run: pytest -q --junitxml=pytest-report.xml
-      - if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: pytest-report
-          path: pytest-report.xml
-          if-no-files-found: error
+      - run: pytest -q
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,4 +15,10 @@ jobs:
           cache: pip
       - run: python -m pip install -e ".[dev]"
       - run: ruff check src tests
-      - run: pytest -q
+      - run: pytest -q --junitxml=pytest-report.xml
+      - if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: pytest-report
+          path: pytest-report.xml
+          if-no-files-found: error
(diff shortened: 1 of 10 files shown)


FlorianPfaff:
Closing as superseded by #4.

This branch introduces a second public observation schema and CLI around already-fused, gauge-relative products. The retained interface in #4 preserves unfused factors, one explicit Sim(3) nuisance gauge per window, gauge Jacobians, and separate conditional versus gauge-marginal covariance; it is also the interface consumed by the newer Bayesian-PhysTwin#3 integration. Maintaining both contracts would duplicate validation/serialization code and leave downstream users with incompatible uncertainty semantics.

The useful causal-export idea here should be reintroduced, if needed, as a focused adapter that selects only causally complete source windows and emits the ObservationFactorBundle schema from #4.

FlorianPfaff:
Closure note: this early artifact-contract branch is superseded by merged PRs #4, #6, and #7, with the production joint-covariance, strict loader, packaging, and Prob4D 0.2.0 hardening completed in PR #11. The merged contracts preserve the useful causal, provenance, covariance, and tamper-detection requirements while avoiding a second incompatible artifact surface.

PR #9, opened 2026-07-26
Export causally sealed portable observation beliefs

FlorianPfaff:
## Summary

Add the missing production bridge from MotionCrafter prediction bundles to the provider-neutral `phys4d.observation_belief` artifact consumed by Bayesian-PhysTwin and independently validated by Causal4D.

## What changed

- add a byte-addressed `ObservationBeliefExportV1` writer matching the existing cross-repository schema and golden digest;
- add a content-addressed fixed metric-gauge anchor bound to the exact first retained prediction window and calibration artifact;
- select causally complete independently decoded windows from manifest metadata before opening prediction payloads;
- recompute relative alignment, gauge estimation/smoothing, overlap disagreement, uncertainty, and prior reliability only on the admitted prefix;
- export conditional metric covariance plus shared low-rank `Sim(3)` gauge factors, avoiding gauge double counting;
- retain association probability and prior reliability as separate fields;
- cap repeated dense evidence with frame-level correlation groups and composite-likelihood weights;
- add `prob4d-create-metric-gauge-anchor` and `prob4d-export-observation-belief` CLIs;
- document the causal, covariance, metric-anchor, and provenance contracts.

## Root cause addressed

Filtering output rows after estimating gauges and uncertainty on a full prediction bundle is not causally safe: post-cutoff overlap frames can alter pre-cutoff transforms, covariance, disagreement, and reliability. The new path admits complete source windows first and performs every statistical operation after that selection.

The source content address deliberately excludes unconsumed future-entry bookkeeping. Appending post-cutoff windows therefore leaves the selected source digest and final observation artifact ID unchanged, and crossing payload files are never opened.

## Metric-anchor boundary

Portable `ObservationBeliefV1` has one coherent seven-dimensional gauge factor per window. A nonzero global metric-anchor covariance would require another coherent factor shared across every window and cannot be represented correctly at the fixed rank used by V1. This exporter therefore requires a fixed external metric calibration. Users who need an uncertain global anchor must use the existing `ObservationFactorBundle` interface instead.

## Validation

Focused regression coverage includes:

- the shared cross-repository golden artifact ID;
- descriptor/array content addressing;
- future-frame and duplicate-row rejection;
- metric-anchor round trip and reference-payload binding;
- fail-closed source-frame validation;
- future payloads that do not exist because they must not be opened;
- exact artifact invariance after appending post-cutoff windows;
- recovery of `J Sigma_g J^T` from the exported low-rank factor;
- rejection of uncertain global anchors on the compact V1 path.

The repository workflow will run Ruff and the complete pytest suite on this draft PR.

## Claim boundary

This PR creates an auditable uncertainty/provenance boundary. It does not claim that Prob4D observations improve a physical twin or that the exported covariance is prospectively calibrated. Relative window gauges are represented by marginal low-rank factors; residual cross-window dependence remains conservatively capped by composite weights pending a future joint sparse gauge posterior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,44 @@ It also records a versioned temporal-lineage contract for MotionCrafter's
 internal sliding windows. Causal consumers can therefore determine the exact
 source-frame bounds and contributing internal windows for any output frame.
 
+### Portable Bayesian observation export
+
+First bind a prefix-only metric calibration to the exact first retained window:
+
+```bash
+prob4d-create-metric-gauge-anchor \
+  outputs/sequence_name/predictions.json \
+  outputs/sequence_name/metric_gauge_anchor.json \
+  --case-id sequence_name \
+  --world-frame-id phystwin-world \
+  --reference-window-id window_0000 \
+  --calibration-artifact outputs/sequence_name/prefix_registration.json \
+  --sim3-vector LOG_SCALE RX RY RZ TX TY TZ
+```
+
+Then export the provider-neutral observation artifact consumed by
+Bayesian-PhysTwin and independently validated by Causal4D:
+
+```bash
+prob4d-export-observation-belief \
+  outputs/sequence_name/predictions.json \
+  outputs/sequence_name/observation_belief.npz \
+  --case-id sequence_name \
+  --causal-frame-stop 134 \
+  --metric-gauge-anchor outputs/sequence_name/metric_gauge_anchor.json \
+  --summary-json outputs/sequence_name/observation_belief_summary.json
+```
+
+The exporter opens only independently decoded windows wholly before the
+exclusive cutoff, then recomputes alignment, gauge estimation, overlap
+disagreement, uncertainty, and residual-independent reliability on that prefix.
+Appending post-cutoff windows cannot change the exported artifact ID.
+Association probability remains separate from prior reliability, and gauge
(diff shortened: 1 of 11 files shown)


FlorianPfaff:
Closing as superseded by merged PRs #6 and #7.

This PR substantially duplicates the causal metric observation-export path now on `main`. It also computes `group_prior_nominal_probability` from a quantile of row-level `prior_reliability` while exporting those same row reliabilities separately, which applies the overlap-derived evidence twice. PR #7 deliberately uses the neutral group prior `1.0` unless an independently calibrated nominal/outlier prior exists.

The separately content-addressed anchor-authoring CLI in this branch is useful and can be proposed later as a small focused addition, but the overlapping exporter should not be merged after #7.

FlorianPfaff:
Closing as superseded. PR #7 was merged with the fixed-anchor compact V1 semantics and neutral group prior. PR #10 carries forward the remaining provider-boundary, immutability, packaging, and CI improvements.

FlorianPfaff:
Closure note: this draft is superseded by merged PR #7 and the later covariance/validation hardening in PR #11. The final implementation preserves causal source admission, metric anchoring, separate reliability terms, content addressing, joint cross-window gauge covariance, and strict artifact validation without retaining this parallel implementation.

PR #15, opened 2026-07-26
Add a portable joint-observation conformance fixture

FlorianPfaff:
## Summary

Add a small, provider-neutral three-window observation fixture that freezes the current Prob4D joint-gauge covariance semantics.

- records one exact content-addressed `ObservationBeliefV1` payload;
- uses one shared joint gauge latent vector across three windows;
- contains nonzero cross-window covariance;
- records causal source lineage, metric anchoring, factor rank, retained covariance trace, and parent-window lineage;
- verifies the producer save/load round trip without requiring MotionCrafter or GPU dependencies.

## Why

Prob4D now exports a shared low-rank root of the joint `Sim(3)` gauge covariance. A compact semantic fixture is needed so Bayesian-PhysTwin and Causal4D can independently verify the same producer meaning, rather than checking only a legacy byte-layout example.

## Validation

The new test reconstructs the artifact from the checked-in JSON fixture, verifies the exact artifact ID, confirms nonzero cross-window covariance and one shared factor group, then validates an NPZ save/load round trip.

The fixture is copied byte-for-byte into the companion Bayesian-PhysTwin and Causal4D PRs.

## Claim boundary

Test and interoperability infrastructure only. This PR does not change the gauge estimator, covariance computation, observation values, or any empirical claim.


FlorianPfaff:
## Cross-repository validation

Companion PRs:

- `FlorianPfaff/Bayesian-PhysTwin#13` consumes and validates the joint-gauge artifact.
- `FlorianPfaff/Causal4D#41` independently validates and binds its causal lineage.

All three repositories contain the identical fixture blob (`1a36d2a89e278bb8dd1c83d3491808b691d0f6fb`) and independently obtain artifact ID `2a1f24acd2dd741155eb5333c92a37f615e1f0578b7f55cb9df0d34c52af8640`.

This PR's complete CI passed: Ruff, stable-provider type checking, Python 3.10/3.12/3.14 suites, provider manifest, distribution builds, and isolated wheel/sdist CLI smoke tests.

PR #16, opened 2026-07-26
Fix the cross-repository joint observation contract

FlorianPfaff:
## Summary

Repair the producer side of the Prob4D → Bayesian-PhysTwin → Causal4D observation boundary.

- version Prob4D-specific observation semantics independently of the generic `ObservationBeliefV1` container;
- declare the production joint Sim(3) tree layout and the explicitly approximate fixed-lag layout;
- bind metric anchors to the exact reference-window payload and calibration artifact;
- distinguish fixed external calibration from anchor covariance propagated through the shared joint factor;
- route the stable CLI and `prob4d.provider_v1` through the portable contract boundary;
- write anchor JSON and observation NPZ outputs atomically, with strict NPZ reload before publication;
- add `prob4d observation create-anchor` / `prob4d-create-metric-gauge-anchor`;
- retain legacy direct core APIs and legacy anchor loading for frozen experiments;
- release the producer fix as Prob4D 0.2.1.

## Compatibility

The generic `phys4d.observation_belief` schema remains version 1. New machine-readable metadata carries `prob4d_observation_contract_version = 2`, the covariance layout, and factor-group semantics. Legacy per-window seven-factor artifacts remain a consumer-side compatibility path.

Coordinated consumer PRs in Bayesian-PhysTwin and Causal4D will independently validate both layouts and reject ambiguous partially migrated artifacts.

## Validation

Locally performed:

- Python compilation of all changed modules and tests;
- focused smoke execution of metric-anchor round trips, contract enrichment, atomic serialization, and joint/legacy/approximate consumer validation with isolated stubs;
- line-length checks matching repository Ruff configuration.

The repository GitHub Actions suite is the authoritative full-package validation because this private repository cannot be cloned into the local execution environment.


FlorianPfaff:
Superseded by #17, which is based on the merged cross-repository conformance fixture and already passes the complete Prob4D CI suite. The useful remaining pieces from this draft—metric-anchor provenance and atomic publication—will be handled against the canonical stream-contract branch rather than maintaining a second contract dialect.

PR #20, opened 2026-07-26
Add content-addressed covariance calibration

FlorianPfaff:
## Summary
- add content-addressed gauge and point covariance calibration artifacts with strict provenance and tamper detection
- make provider spatial-cluster covariance fallbacks fail closed by default and record explicit fallback use
- integrate calibrated exports with causal stream contract v2 while keeping approximate fixed-lag output unbound
- harden immutable `Sim3`, alignment-result, and structured-covariance contracts
- document claim-bearing calibrated export and add regression coverage

## Tests
- 27 targeted tests passed locally
- `python -m compileall` passed
- `ruff` and `mypy` were unavailable in the local environment

Prospective real-data covariance-calibration and downstream physical-improvement claims remain explicitly false in the provider manifest.


FlorianPfaff:
### Conflict resolution completed

Current `main` was integrated with an explicit audited resolution:

- content-addressed gauge and point covariance calibration retained;
- fail-closed spatial-cluster covariance and explicit fallback audit retained;
- metric-anchor provenance and covariance propagation retained;
- causal observation stream contract v2 retained for sequential exports;
- calibration metadata is attached before strict stream binding;
- approximate fixed-lag output remains explicitly unbound;
- temporary diagnostic workflows are absent from the final tree.

Validation passed: Ruff, mypy, provider-manifest emission, full pytest on Python 3.10/3.12/3.14, and wheel/sdist build plus `twine check`. The downstream isolated-install smoke jobs were still awaiting hosted-runner allocation at merge time.

PR #21, opened 2026-07-26
Fail closed on invalid covariance matrices

FlorianPfaff:
## Summary

Replace silent inversion/eigenvalue clipping of materially indefinite covariance matrices with scale-aware, fail-closed validation.

## Changes

- add shared covariance eigendecomposition, inverse, and log-determinant helpers;
- require finite, symmetric, positive-semidefinite covariance input;
- tolerate only floating-point-scale negative eigenvalues before regularization;
- route independent Gaussian fusion and covariance intersection through the validated inverse;
- check covariance-intersection determinant signs explicitly;
- make uncertainty diagnostics and sequence NLL/coverage reject invalid covariance instead of repairing it silently;
- stop silently dropping non-finite covariance rows from diagnostics;
- fix the legacy flow metric to intersect deformation support with valid geometry and finite flow values;
- add adversarial regression tests for indefinite, asymmetric, tiny-negative, and invalid-flow inputs.

## Repair

The failed self-applying patch and workflow wrapper were removed. This PR now contains only the four direct production/test changes rebased onto the merged covariance-calibration work from #20.


FlorianPfaff:
### Broken patch carrier repaired

The failed self-applying patch/workflow mechanism has been removed. The PR now contains exactly four direct changes:

- `src/prob4d/covariance.py`
- `src/prob4d/fusion.py`
- `src/prob4d/metrics.py`
- `tests/test_covariance_validation.py`

The implementation fails closed on non-finite, asymmetric, and materially indefinite covariance matrices; preserves only scale-aware floating-point regularization; checks covariance-intersection determinants; and excludes invalid geometry/flow from the legacy flow metric.

Validation passed: Ruff, mypy/provider contract, and the complete pytest suite on Python 3.10, 3.12, and 3.14. Distribution packaging was still awaiting hosted-runner allocation at merge time.

PR #24, opened 2026-07-27
Add order-invariant gauge graph inference

FlorianPfaff:
## Summary

- add deterministic frame/tile/depth-stratified sampling for dense overlap alignment;
- preserve frame-by-spatial-tile sandwich covariance and fail closed when fewer than eight independent clusters survive;
- add manifold-local generalized covariance intersection for several uncertain `Sim(3)` candidates without pairwise order dependence;
- add orientation-independent overlap-edge content IDs and a deterministic global minimum-uncertainty-volume spanning tree;
- propagate the selected tree into one full cross-window gauge covariance;
- add held-out loop-closure residual and NIS diagnostics without silently treating redundant edges as independent updates;
- cover constraint permutations, near-π rotations, graph connectivity, covariance propagation, loop faults, exact sampling budgets, depth coverage, and fallback semantics;
- document the estimator as an additive V2 ablation path rather than changing the frozen portable observation stream.

## Why

The current sequential gauge initializer applies pairwise covariance intersection in candidate iteration order, while the strict observation export selects one parent greedily for each new window. Dense overlap alignment also uses an unstratified random subset when correspondence count exceeds the configured cap. These choices can make diagnostics sensitive to ordering or sampling while discarding useful graph structure.

This PR provides deterministic alternatives while preserving all existing defaults, artifact schemas, stream-contract versions, and frozen experiment semantics.

## Validation

Focused local validation against the current NumPy core:

```text
10 passed
```

The focused suite exercises generalized-CI permutation invariance, sequential constraint-order invariance, deterministic global tree selection, full joint covariance, loop-fault detection, SO(3) branch handling, stratified sampling determinism, spatial/depth coverage, exact `Sim(3)` recovery, and fail-closed covariance fallback.

GitHub Actions remains authoritative for Ruff, the complete Python 3.10/3.12/3.14 suite, distribution builds, and installed wheel/sdist smoke tests.

## Claim boundary

This is an estimator and diagnostic implementation, not a held-out performance result. Generalized CI handles unknown correlation among parent candidates; selected tree-edge errors remain a first-order covariance model, and non-tree NIS values are explicitly diagnostic. Promotion into the production observation stream still requires the registered sequence-family-held-out Prob4D and downstream Bayesian-PhysTwin gates.

### DIFF
--- a/docs/gauge-estimator-v2.md
+++ b/docs/gauge-estimator-v2.md
@@ -1,0 +1,101 @@
+# Gauge Estimator V2: order-invariant graph inference
+
+Prob4D's portable causal observation stream deliberately retains the frozen
+`sequential_joint_spanning_tree_v1` behavior. This document describes an additive
+estimator and alignment path for controlled ablations before any production
+contract is changed.
+
+## Motivation
+
+The original sequential estimator processes an explicit temporal window order,
+but it fuses several already-initialized parent candidates through pairwise
+covariance intersection. Pairwise CI is not associative, so permuting otherwise
+identical constraints can change the result. The strict observation exporter also
+selects one parent for each new window greedily and does not use non-tree overlap
+edges except as recorded metadata.
+
+Dense overlap alignment has a related sampling problem. When more than the
+configured maximum number of correspondences are available, an unstratified
+random subset can retain many redundant pixels while omitting a small region that
+provides useful scale, rotation, or depth leverage.
+
+## Additive APIs
+
+`prob4d.gauge_graph` provides:
+
+- `OrderInvariantSequentialGaugeEstimator`, which maps all parent candidates to a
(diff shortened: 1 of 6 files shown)


FlorianPfaff:
Closing as redundant after merging #23. Both PRs implement order-invariant manifold-local multi-parent Sim(3) covariance intersection, but #23 fixes the active `SequentialGaugeEstimator` with a substantially smaller, fully integrated patch. Merging this PR as well would leave two independent CI implementations (`_gauge_ci.py` and `gauge_ci.py`) and a second, largely disconnected estimator stack. The graph-tree, loop-diagnostic, and stratified-alignment ideas remain available on this branch for a future focused PR with an explicit provider/CLI integration and held-out promotion gate.

PR #26, opened 2026-07-27
Preserve provider information mass and make fusion order-invariant

FlorianPfaff:
## Summary

- define provider-owned per-row correlation-group information mass with auditable raw, unique, and effective counts
- rank joint Sim(3) gauge modes by sampled observation-space variance rather than unit-dependent raw coordinate trace
- jointly fuse every frame/pixel contributor using associative multi-input formulas and generalized covariance intersection
- canonicalize windows by immutable ID and add permutation, duplication, unit-invariance, and manifest regression tests
- document the new provider/consumer semantics

## Validation

The exact staged patch passed 171 source-distribution tests locally. A temporary self-cleaning workflow applies the tested bytes to this branch, runs Ruff, mypy, and the complete repository suite, then removes itself and the patch staging files. Normal repository CI will run on the resulting implementation commit.

## Claim boundary

This is an estimator and interoperability improvement. It does not create a prospective covariance-calibration or physical-twin-benefit claim.

### DIFF
--- a/.github/agent-patches/prob4d-improvements.patch.part-00
+++ b/.github/agent-patches/prob4d-improvements.patch.part-00
@@ -1,0 +1,305 @@
+diff -ruN '--exclude=__pycache__' '--exclude=.pytest_cache' prob4d_clean/README.md prob4d_current/README.md
+--- prob4d_clean/README.md	2026-07-27 06:18:23.000000000 +0000
++++ prob4d_current/README.md	2026-07-27 06:35:02.249634307 +0000
+@@ -124,6 +124,7 @@
+ 
+ See [the experiment protocol](docs/experiment-protocol.md) for estimator and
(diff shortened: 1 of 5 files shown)


FlorianPfaff:
Closing as redundant/incompatible after #25 was merged. The provider-owned per-row information mass and unit-stable Sim(3) rank-reduction goals are already implemented on `main`, with the exact semantic identifier consumed by Bayesian-PhysTwin PR #47. This PR reimplements those contracts under different identifiers and from the pre-#23/#25 base, so resolving it would create duplicate semantics rather than a clean conflict fix. It also bundles a separate 1,000+ line joint-fusion rewrite into temporary patch/workflow staging. The generalized multi-input CI and batch-fusion portion may be valuable, but it should return as a focused PR rebased on current `main`, preserving the merged provider contract and carrying dedicated numerical/performance validation.

PR #34, opened 2026-07-28
Preserve joint gauge covariance in unfused factors

FlorianPfaff:
## Summary

- add `JointObservationFactorBundle` schema v4 with one ordered full cross-window `Sim(3)` gauge covariance;
- validate covariance shape, finiteness, symmetry, positive semidefiniteness, gauge order, and equality of every diagonal block with its `GaugeEstimate` marginal;
- preserve the full joint prior when factors are stacked, so explicit-gauge consumers retain shared anchor and upstream-window uncertainty;
- serialize the joint covariance in the checksum-bound NPZ payload and bind its semantics and gauge order in the manifest;
- retain provider-v1 `ObservationFactorBundle` schema v3 and its historical block-diagonal stacking behavior unchanged;
- advertise schema v4 and `joint_observation_factor_gauge_covariance` only through provider v2;
- document the conditional-versus-marginal covariance boundary and legacy migration semantics.

## Why

Schema v3 stores only marginal gauge covariances. Its stack therefore constructs a block-diagonal nuisance prior, which discards cross-window covariance induced by a shared metric anchor and sequential relative-gauge constraints. Using that representation for new multi-window explicit-gauge inference can overstate independent information.

A silent reinterpretation of schema v3 would break frozen reproduction. Schema v4 makes the stronger covariance claim explicit and fail-closed while preserving exact v1 compatibility.

## Validation

Local validation:

- `31 passed`: observation-factor, provider-v1/v2 manifest, and provider-attestation focused tests;
- `231 passed`: all tests available in the downloaded source distribution;
- `git diff --check` passed;
- source and tests compile successfully.

GitHub Actions run 347 completed successfully:

- Ruff;
- mypy stable-provider checks;
- runtime revision attestation and provider-manifest emission;
- complete test suites on Python 3.10, 3.12, and 3.14;
- wheel and source-distribution build plus `twine check`;
- isolated wheel and sdist installation;
- grouped and legacy CLI smoke tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to Prob4D are documented here.
 
 ### Added
 
+- `JointObservationFactorBundle` schema v4 for provider-v2 explicit-gauge
+  consumers. It serializes one ordered full cross-window `Sim(3)` covariance,
+  validates every marginal block against its `GaugeEstimate`, and preserves the
+  complete nuisance prior when factors are stacked.
+- Provider-v2 capability and schema declarations for joint observation-factor
+  gauge covariance while provider v1 continues to advertise frozen schema v3.
 - `prob4d.provider_v2` as a safe-by-default Python surface with distinct
   exploratory and claim-bearing export functions.
 - Strict prediction/calibration compatibility checks covering MotionCrafter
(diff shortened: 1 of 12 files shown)


FlorianPfaff:
Superseded by #35, which has now been merged as fd6b49eea7347b8f1d9f58e951bcdcaeca83762e. The merged implementation keeps one versioned bundle type with explicit `joint-cross-window` versus `marginal-blocks-only` semantics and preserves the frozen provider-v1 schema-v3 boundary.

PR #72, opened 2026-08-04
Add an authoritative fail-closed quality gate

FlorianPfaff:
## Summary

Replace the blocked self-modifying workflow with a normal read-only authoritative quality gate.

- run Ruff through `tee` with Bash `pipefail`, so diagnostics cannot turn a failing linter into a green step;
- run stable-interface mypy through `tee` with the same fail-closed behavior;
- constrain only this Python-3.10-targeted typing environment to compatible NumPy stubs `<2.3`, leaving the runtime compatibility matrix unchanged;
- cover the claim-bearing factor envelope, provider-v2 factor facade, sparse explicit-gauge stack, and source-only diagnostics in stable-interface typing;
- retain immutable Action pins, non-persisted checkout credentials, diagnostic artifact upload, and read-only workflow permissions;
- add executable policy tests proving that the authoritative workflow has no `continue-on-error`, uses `pipefail` for every diagnostic pipeline, covers the current stable surfaces, and rejects unsafe fixture pipelines;
- retain the focused source typing corrections already identified by the newly authoritative mypy run.

## Why

The former Tests quality job pipes Ruff and mypy through `tee` and has historically reported success after the producer command failed. The previous repair attempted to rewrite and push its own workflow, which GitHub correctly held for approval. This revision adds a conventional, independently required failing job instead of requesting repository-write permissions.

This changes CI correctness and static maintenance only. It does not alter provider, calibration, observation, factor, fusion, evaluation, BayesianPhysTwin, or Causal4D numerical semantics.

## Validation

The new `Fail-closed quality` workflow is authoritative for Ruff and stable-interface mypy. The normal complete Python, minimum-dependency, build, wheel, sdist, and installed-command matrix also remains required. The PR stays draft until the merged-base runs are green.


FlorianPfaff:
Closing as superseded by #76. This draft still carries unrelated stacked gauge-graph and factor-surface changes, whereas #76 is the explicitly stated replacement rebased onto current `main`. No unique scientific or provider behavior from this branch is being discarded by the closure.

PR #76, opened 2026-08-04
Make quality checks fail closed

FlorianPfaff:
## Summary

- make Ruff and mypy diagnostic pipelines propagate the producer exit status through `tee` by enabling `pipefail`;
- pin the Python-3.10-targeted mypy job to compatible NumPy stubs while retaining the unconstrained runtime matrix;
- include the provider-v2 factor-bundle surface in stable mypy and import checks;
- add a workflow policy test that rejects diagnostic pipelines which omit `pipefail`;
- apply the type annotations and Ruff cleanups exposed once the previously masked checks become authoritative.

## Validation

A clean checkout of current `main` with this exact source/test diff and the proposed `tests.yml` applied passed:

- Ruff: all checks passed;
- mypy: no issues in 19 stable source files;
- pytest: **439 passed**;
- the workflow-policy test confirms that every diagnostic pipeline using `tee` enables `pipefail`.

The PR diff contains only the final workflow, source, and test fixes. All temporary self-applying workflows and patch payloads used to work around GitHub workflow-file token restrictions have been removed.

## Scope

CI correctness, typing, and style only. No estimator, provider schema, artifact identity, calibration semantics, BayesianPhysTwin behavior, or scientific claim changes.

Supersedes #72.


FlorianPfaff:
Closing as superseded by merged PR #72 (`60cadacf5bbdfff71a40b995e3b178993fe6ab6f`). Current `main` already contains the fail-closed Ruff/mypy workflow, compatible typing environment, stable-interface coverage, and associated type/style fixes. This branch also retains obsolete self-applying patch files, so merging it would reintroduce staging machinery rather than add value.

FlorianPfaff:
Closing as superseded by merged PR #72. I initially reviewed an older head that still contained one-shot application and embedded patch payloads; the branch was subsequently rewritten to a clean 19-file diff, so that staging-specific observation no longer applies to the final head. Rechecking the rewritten branch, its source/type corrections and authoritative fail-closed Ruff/mypy behavior are already on `main` through #72. The only material delta is a second, overlapping repair of the legacy `Tests` quality job plus related policy assertions. That cleanup should be proposed as a small current-main follow-up if still desired, rather than merging this conflicted 14-commit branch whose description claims to supersede an implementation that is already merged.

Issue #82, opened 2026-08-04
Control worst-group cycle-guard false fallback with source-only finite-sample calibration


FlorianPfaff:
## Motivation

The preregistered uncertainty-normalization experiment in PR #81 retained detection of all `89/89` injected precise-looking biased edges and reduced the worst clean false-fallback rate from 29.7% to 13.3%. It nevertheless failed the frozen absolute ceiling of 10%, so the normalized guard remains experimental.

The clean target exceedance rates were:

| Registered clean regime | Raw guard | Normalized guard |
| --- | ---: | ---: |
| independent | 29.7% | 9.4% |
| correlated | 0% | 13.3% |
| highly correlated | 0% | 7.8% |

Normalization largely equalized the mean statistic across regimes, but a single pooled empirical threshold did not control the group-specific tails with 48 calibration trials.

## Scientific hypothesis

A finite-sample threshold calibrated to control the worst predeclared source-uncertainty group can reduce the normalized guard's worst clean false-fallback rate to at most 10% without materially reducing mild or strong biased-edge detection.

## Candidate calibration strategies

Evaluate transparent source-only candidates before target access:

1. finite-sample conformal order statistics with explicit coverage level and randomized/nonrandomized tie policy;
2. the maximum of per-stratum conformal thresholds over predeclared source-uncertainty strata;
3. a monotone quantile model using only source covariance diagnostics, with leave-one-group-out calibration;
4. a conservative tolerance-bound threshold as a sample-size control.

A stratum cannot use the synthetic generator's hidden correlation parameter in the method. It must be derivable from observable source metadata, such as direct/path propagated uncertainty scale, effective correspondence count, residual scale, covariance anisotropy, or CI effective-edge count. The mapping from source features to strata must be frozen on development/calibration trials.

## Required information boundary

The method and all thresholds must use only:

- alignment transforms and source-side covariance;
- overlap support/correspondence diagnostics;
- predeclared calibration groups and seeds.

They must not use target truth, target error, BayesianPhysTwin innovations, association probability, or downstream physical residuals.

## Frozen comparison

Use fresh calibration and target seed ranges, retain the six scenarios and outlier semantics from `prob4d-cycle-guard-normalization-v1`, and compare:

- raw-displacement guard;
- pooled normalized guard from PR #81;
- each new group-robust normalized candidate;
- unguarded full-joint graph;
- exact production tree.

## Primary endpoints and gate

All conditions must pass:

[...]

PR #86, opened 2026-08-04
Evaluate finite-sample group-robust cycle calibration

FlorianPfaff:
## Scientific hypothesis

Issue #82 asks whether source-only finite-sample calibration can reduce the uncertainty-normalized cycle guard's worst clean false-fallback rate to at most 10% without sacrificing biased-edge detection.

## Frozen primary method

The primary method was fixed before target evaluation:

1. use 96 fresh clean development trials to freeze tertiles of the observable `median_cycle_minkowski_uncertainty_scale`;
2. assign 192 fresh clean calibration trials to those observable strata;
3. fit a nonrandomized 95% split-conformal upper threshold in each stratum using order index `ceil((n+1)*(1-alpha))`;
4. use the **maximum** of the three stratum thresholds as the primary global guard threshold.

The hidden synthetic correlation label is retained only for simulation audit and is never supplied to the feature map, stratum assignment, threshold, or guard.

## Pre-target audit

No fresh target outcomes have been evaluated. Two pre-target checks failed closed:

- Ruff found and removed one unused import.
- A focused regression disproved the provisional assumption that the observable Minkowski scale is invariant to translation-edge injection. Because the feature's point Jacobian is transform-dependent, the adversarial translation can slightly change the feature and a secondary stratum-specific control's assignment.

The branch now records the pre- and post-injection feature and their difference. The protocol explicitly states:

- the feature may change under translation injection;
- the **primary maximum-stratum threshold is independent of target stratum assignment**;
- only the secondary stratum-specific control depends on that assignment.

No target seed, method, threshold rule, endpoint, gate, or claim boundary was altered. The branch is synchronized with current `main`.

## Locked controls

- production joint spanning tree;
- unguarded full-joint graph;
- raw pooled empirical 0.95 guard;
- PR #81-style pooled empirical normalized guard;
- pooled conformal normalized guard;
- observable-stratum-specific conformal guard;
- primary maximum-stratum conformal guard.

## Fresh evaluation

- development seed start `380260804`, 96 trials;
- calibration seed start `480260804`, 192 trials;
- target seed start `571260804`, 128 trials per each of six scenarios, 768 total;
- 2,000 paired target-trial bootstrap resamples;
- exact complete-tree transform and covariance equality required on every fallback.

## Registered gate

Every primary condition must pass:

- strong biased-edge detection at least 0.95;
- strong detection no more than 0.05 below the previous normalized guard's 1.00;
- mild biased-edge detection at least 0.90;
- mild detection no more than 0.05 below the previous normalized guard's 1.00;
- worst registered clean false fallback at most 0.10;
- exact production-tree fallback on every rejection.

The workflow treats exit code 3 as a completed negative study and uploads its complete evidence.

## Information and claim boundary

The method may use alignment transforms, source-side covariance, and overlap diagnostics. It may not use target truth/error, hidden correlation as a method input, BayesianPhysTwin innovation, association probability, or downstream physical residual.

Even a positive result remains controlled synthetic source-admission evidence only. It does not establish held-out physical-object provider competence, BayesianPhysTwin benefit, harmful-update control, or Causal4D benefit.


FlorianPfaff:
Closing in favor of #87.

This branch contains a second finite-sample threshold implementation (`FiniteSampleConformalThreshold`) for the same normalized cycle score. Compared with #87's reusable `FiniteSampleUpperThreshold`, it has weaker exact-type validation, no content-addressed calibration-score digest, and different zero-threshold semantics. Merging both would create two public/internal contracts for the same order statistic and duplicate the pooled conformal study and CLI infrastructure.

The observable-stratum idea is not rejected, but it should be a follow-up built on the single strict threshold API from #87 and justified by #87's independently seeded pooled result. The branch and frozen design remain available if that result shows residual subgroup heterogeneity worth testing.

FlorianPfaff:
Closing as redundant with merged #87.

#87 now supplies the single strict finite-sample threshold contract and has completed the frozen 96/768 study with all 10 registered criteria passing: 104/104 injected biased edges detected, worst registered clean false fallback 7.03%, and exact production-tree fallback semantics. That directly resolves the motivating ≤10% clean-fallback question without observable-stratum tuning.

This branch adds a second threshold implementation and substantially more development/calibration machinery for a subgroup-robust variant whose primary purpose is no longer unmet. The stratum-specific method also depends on an observable feature already shown to move under the injected perturbation. Retain the design as a possible future heterogeneity study only if independent evidence reveals a subgroup failure after #87; it should not remain an open parallel implementation now.

Issue #49, opened 2026-07-31
Run a support-feasible held-out real provider and BayesianPhysTwin gate


FlorianPfaff:
## Objective

Test one genuinely independent real provider on complete physical object/session groups, separating provider competence from downstream guarded BayesianPhysTwin value. Controlled synthetic graph and Prob4D-to-BayesianPhysTwin results are mechanism evidence only; they do not establish real-provider competence or unseen-object benefit.

## Stage 0: support feasibility before residual outcomes

Before opening source residuals or any target outcome, publish a content-addressed `ProviderSupportFeasibilityV1` decision for the exact provider/model/runtime identity. It must bind:

- complete development, calibration, target, and excluded object/session rosters;
- representation, units, metric-frame prior, causal cutoff, windows, seeds, and immutable model set;
- required streams, identities, masks, timestamps, camera geometry, and physical-query mapping;
- minimum complete-stream support and technical-failure policy; and
- exact unsupported-case fallback.

If complete-stream support fails, stop that provider version and report the negative feasibility result. Do not fit covariance, inspect target outcomes, substitute partial streams, or rescue it with a different graph/anchor/model under the same identity.

## Frozen comparison arms

Select at most one Prob4D candidate using development/calibration units only. Score exactly these main arms on the same held-out units:

1. unchanged physical fallback;
2. simple direct visual observation or `last_residual` comparator;
3. one source-selected Prob4D candidate with explicit complete joint gauge uncertainty;
4. exact physical fallback for every unsupported, invalid, or rejected candidate.

Metric-assisted variants must be labelled sensor-assisted and may not be compared as though they used the same information as an unanchored arm. Additional graph, association, semantic, or covariance variants are out of scope unless this frozen result localizes a specific failure that requires a new protocol.

## Freeze before target access

Bind exact Prob4D, BayesianPhysTwin, observation-provider, model, environment, and numerical identities; source/calibration-only reliability and covariance calibration; the BayesianPhysTwin guard; horizons; physical queries; group weighting; bootstrap/randomization seeds; technical failures; and exact fallback. Frames, points, tracks, views, and taxels are nested observations, not independent groups.

### Provider competence

Report by independent object/session and horizon:

- support and technical-failure accounting;
- point/endpoint or valid identity-based track error;
- seam/drift diagnostics where applicable;
- 50/90/95% coverage, proper score, normalized NEES, and full covariance width;
- identity retention/precision and selective risk; and
- worst-group coverage shortfall.

[...]

PR #92, opened 2026-08-04
Add content-addressed joint-gauge tracklet evidence with covariance-safe ranking

FlorianPfaff:
## What changed

- add `CausalTrackletArtifactV1`, binding every tracklet array, nested metadata, prediction-manifest identity, exact producer revision, and builder configuration into deterministic SHA-256 identities;
- add correlated residual-covariance diagnostics for cross-window material candidates using the complete joint `Sim(3)` gauge prior, including `P_lr` and `P_rl` cross terms;
- preserve covariance-independent isotropic geometric compatibility for mutual-best ranking inside the joint-gauge evidence path, so a candidate cannot improve merely by becoming wider and less informative;
- bind tracklet-producer and association-implementation revisions separately;
- state explicitly that conditional point cross-window covariance is unavailable and assumed zero;
- reject unordered/noncanonical gauge-ID containers;
- add adversarial regressions for covariance-inflation ranking, content sensitivity, immutable metadata, noncanonical provenance, positive/negative gauge correlation, PSD failures, source-revision mismatch, and execution-chunk invariance;
- document the mathematical semantics and downstream promotion boundary.

## Mathematical boundary

For residual `r = p_left - p_right`, the gauge contribution is

`J_l P_ll J_l^T + J_r P_rr J_r^T - J_l P_lr J_r^T - J_r P_rl J_l^T`.

The transformed conditional point covariances are added exactly once. Their cross-window conditional covariance is not available from the current producer and is content-bound as `assumed-zero-unavailable-v1`; `joint-cross-window-gauge-v1` must not be interpreted as complete residual dependence.

Joint covariance determines the reported normalized-residual diagnostic. Mutual-best ranking, margins, and minimum compatibility retain the bounded covariance-independent geometric score. A covariance-aware admission policy remains a separately versioned, source-calibrated follow-up rather than a silent semantic change.

## Clean integration

The branch was rebuilt directly on current `main` after #94, #95, #96, and #103 merged. Final head `0129928330124155c5d1fab47d20ef4754117d08` contains exactly four additive files:

- read-only focused workflow;
- mathematical/provenance documentation;
- the joint-gauge evidence module;
- focused adversarial tests.

It does not modify or restore the foundational schema-v2 tracklet/association implementation.

## Validation

All authoritative lanes passed on the clean head:

- focused provenance/ranking/dependence workflow;
- Ruff, Ruff-format, and mypy for the evidence layer;
- joint-gauge, cross-window, and causal-tracklet focused regressions;
- fail-closed quality workflow;
- full Python 3.10–3.14 suite;
- exact declared dependency floor;
- wheel and source-distribution build/audit;
- isolated installed wheel/sdist and CLI smoke tests.

## Scientific boundary

This is an experimental evidence surface. It does not rewrite provider-v2 point IDs, calibrate a global identity posterior, establish provider competence, or establish BayesianPhysTwin/Causal4D benefit. Promotion remains the frozen object/session-held-out gate in #49 with exact physical fallback.

### DIFF
--- a/.github/workflows/joint-gauge-tracklet-evidence.yml
+++ b/.github/workflows/joint-gauge-tracklet-evidence.yml
@@ -1,0 +1,85 @@
+name: Joint-gauge tracklet evidence
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/joint-gauge-tracklet-evidence.yml"
+      - "docs/cross-window-tracklet-evidence.md"
+      - "src/prob4d/cross_window_tracklet_evidence.py"
+      - "src/prob4d/cross_window_tracklets.py"
+      - "src/prob4d/causal_tracklets.py"
+      - "src/prob4d/observation_factors.py"
+      - "src/prob4d/sim3.py"
+      - "tests/test_cross_window_tracklet_evidence.py"
+      - "tests/test_cross_window_tracklets.py"
+      - "tests/test_causal_tracklets.py"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: joint-gauge-tracklet-evidence-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  PIP_DISABLE_PIP_VERSION_CHECK: "1"
+  PYTHONUNBUFFERED: "1"
+
+jobs:
+  focused-contract:
+    name: Provenance, ranking, and dependence contract
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Set up Python 3.12
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+          cache: pip
+          cache-dependency-path: pyproject.toml
+      - name: Install compatible development environment
+        run: |
+          python -m pip install -e ".[dev]"
+          python -m pip install "numpy>=1.24,<2.3"
+      - name: Check focused source quality
+        run: |
+          python -m ruff check \
+            src/prob4d/cross_window_tracklet_evidence.py \
+            tests/test_cross_window_tracklet_evidence.py
+          python -m ruff format --check \
+            src/prob4d/cross_window_tracklet_evidence.py \
+            tests/test_cross_window_tracklet_evidence.py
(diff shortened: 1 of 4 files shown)


FlorianPfaff:
## Pre-merge scientific/provenance corrections

The joint cross-window covariance algebra is correct, but the current association score has one promotion-critical failure mode:

### 1. Covariance inflation can improve the mutual-best rank

`normalized_rms` contains only the Mahalanobis quadratic and the compatibility score is `support_fraction * exp(-0.5 * normalized_rms**2)`. For equal residuals/support, increasing residual covariance lowers the quadratic and therefore raises compatibility. A highly uncertain or anti-correlated gauge pair can consequently outrank a more precise candidate solely because its predictive covariance is wider.

The existing hard Euclidean gate limits the damage but does not remove the ranking inversion within that gate. Before merge, use one of the following explicitly versioned semantics:

- include a covariance-volume term (for example the averaged Gaussian log-score/log-determinant) in the ranking;
- use covariance only for a calibrated admissibility gate while ranking admitted candidates by an uncertainty-independent geometric score; or
- retain the current score only as a labelled diagnostic and prevent it from producing admitted links.

Add an adversarial test with equal support and residual magnitude but different covariance volume. The wider candidate must not become mutual-best merely because it is less informative.

### 2. The dependence label must state the missing conditional cross-covariance

The formula uses the complete joint gauge blocks, but it still adds the two conditional point covariances under zero cross-window conditional point covariance. Because overlapping windows share source frames/backbone predictions, `joint-cross-window-gauge-v1` can otherwise be misread as complete residual dependence. Bind a field such as `conditional_point_cross_covariance_semantics: assumed-zero-unavailable-v1` into the result identity and claim boundary.

### 3. Bind producer and association implementation revisions separately

The result's `source_revision` currently proves only that the two tracklet artifacts carry the same producer revision. It does not bind the revision of the association implementation that generated `result_id`. Add an exact `association_revision`/`implementation_revision` field, or rename the existing field to `tracklet_producer_revision` and require the execution manifest to bind the association revision explicitly.

### 4. Reject unordered `gauge_ids`

`_strict_string_tuple` accepts any iterable despite the `Sequence` annotation. A `set[str]` is therefore accepted and can change gauge order/content identity across hash seeds. Enforce an actual ordered sequence (excluding strings/bytes) and add a set-rejection regression.

These corrections do not require changing provider-v2 or the covariance formula. They make the experimental evidence deterministic and prevent uncertainty magnitude from masquerading as material-identity evidence.

FlorianPfaff:
Queue review: keep this as the dependency root for #94 and #95. It is mergeable at the Git level, but the current diff still contains both hardening scripts and temporary hosted/self-hosted publisher workflows, so it is not yet an ordinary product PR and must not be merged in this state. Required order remains: publish/validate clean #92, then rebase and validate #94, then rebase and validate #95.

PR #94, opened 2026-08-04
Harden causal tracklet and cross-window association contracts

FlorianPfaff:
## What changed

- reject Boolean, floating-point, string, and lossy array aliases before identifiers, indices, counts, probabilities, and public settings are normalized;
- require every seed row to have both valid and source-side deform support, preventing background-only singleton tracks when `minimum_track_length=1`;
- keep target-frame deform-mask handling explicit through frozen `target_deform_mask_policy`: compatible `allow` semantics remain available, while `require` is separately auditable;
- record target-mask terminations separately from missing candidates and bind the policy into tracklet metadata and reports;
- expose read-only `geometric_mean_association_probability` as a length-neutral comparison diagnostic without changing cumulative survival semantics;
- validate cross-window candidates, links, and complete results during direct construction, including unique identities, one-to-one links, domain complements, candidate/link consistency, and disposition counts;
- advance the portable association descriptor to schema version 2 and correct `evaluated_track_pair_count` so it counts only pairs producing a complete score;
- use explicit NumPy `TypeAlias` declarations and strict tuple iteration under the repository’s Ruff/mypy contract;
- expand documentation and adversarial regressions across producer and association boundaries.

## Compatibility and scientific boundary

- source deform support is always required for a material-track seed;
- `target_deform_mask_policy="allow"` does not reinterpret target masks as material-membership labels;
- `target_deform_mask_policy="require"` is available only for explicitly registered producer/experiment semantics;
- geometric-mean support is an uncalibrated comparison diagnostic, not source reliability or posterior responsibility;
- result identities intentionally change where schema-v2 accounting or stricter invariants change the descriptor;
- no provider-v2 claim-bearing identity, BayesianPhysTwin acceptance decision, calibration method, physical result, or Causal4D result changes.

## Validation

Final head `94a2d44cee1f7d19a2d0ec697acd92dddaa9e119` passed:

- focused tracklet and cross-window association contract workflow;
- fail-closed Ruff and stable-interface mypy workflow;
- complete Python 3.10–3.14 test matrix;
- exact NumPy 1.24.0 dependency floor;
- wheel and source-distribution build and audit;
- isolated installed wheel/sdist and CLI smoke tests.

Temporary repair and formatting workflows/scripts were removed. This PR is the required foundation for #92 joint-gauge evidence and #95 append-only material-identity hypotheses.


FlorianPfaff:
## Merge-order boundary

This PR overlaps `cross_window_tracklets.py`, its descriptor schema, tests, and documentation with PR #92. PR #92 now carries the promotion-critical covariance-safe ranking correction: supplied covariance changes the reported normalized residual diagnostic but cannot increase mutual-best compatibility merely by becoming wider and less informative.

Please do not merge this branch at its current base (`2d6df37...`). Rebase it after #92, retain the schema-v2 constructor/accounting invariants from this PR, and preserve all of the following from #92:

- `RANKING_SEMANTICS = isotropic-geometric-mutual-best-covariance-diagnostic-v1` in the base result identity;
- covariance-independent compatibility ranking in `associate_cross_window_tracklets`;
- the covariance-inflation adversarial regression;
- the explicit documentation that covariance-aware admission requires a separately versioned, source-calibrated rule.

After that rebase, rerun the complete version/minimum-dependency/package matrix and verify that schema-v2 descriptors bind the ranking semantics. This ordering avoids a clean conflict resolution accidentally restoring Mahalanobis-only ranking.

PR #95, opened 2026-08-04
Add append-only material-identity hypothesis streams

FlorianPfaff:
## What changed

- add `prob4d.material_identity_stream`, an experimental content-addressed stream for carrying sparse cross-window tracklet hypotheses through a causal sequence;
- preserve every endpoint as window-local `(window_id, track_id)` rather than rewriting provider observation identities;
- retain every scored candidate, pairwise mutual-best decision, source support and geometry, complete rejection accounting, and originating schema-v2 association `result_id`;
- admit exactly one new target window per update, with every edge directed from an already admitted source window to that target;
- require sorted unique sources, one common causal cutoff, nondecreasing cutoffs, contiguous update indices, and an exact previous-update hash chain;
- support several prior windows for one target while retaining cross-source conflicts as explicit alternatives rather than silently merging them;
- require canonical root provenance: exact `owner/name` repository identity and lowercase 40-character Git SHA; branch names, tags, abbreviated revisions, outer whitespace, and invalid tuple aliases fail closed;
- add strict duplicate-key/unknown-field loading, immutable nested metadata, atomic persistence, documentation, focused CI, and adversarial tests.

## Core invariants

- pairwise results are oriented `left = admitted source`, `right = new target`;
- a target window is admitted once;
- future windows cannot appear as sources;
- cycles and retroactive edge insertion are impossible by construction;
- prior update objects and IDs remain unchanged after append;
- selected pairwise links remain ranking/gate evidence, not calibrated global identity probabilities;
- no global point ID, connected-component label, or provider-v2 identity rewrite is emitted.

## Validation

After merging schema-v2 foundation #94, head `99a5228aeaa96d0216b483ec3b76a10f177cf2a3` passed:

- clean merge and focused schema-v2 integration validation;
- focused material-identity workflow;
- fail-closed Ruff and stable-interface mypy workflow;
- complete Python 3.10–3.14, dependency-floor, build, wheel/sdist, and installed-artifact workflow.

The final PR contains only source, tests, documentation, and one read-only hosted validation workflow. All temporary repair and refresh machinery is removed.

## Compatibility and scientific boundary

This is an additive experimental contract. It does not change claim-bearing provider-v2 export, pairwise association behavior, BayesianPhysTwin update acceptance, calibration mathematics, or Causal4D behavior. Pairwise-selected hypotheses remain uncalibrated and must not be interpreted as global material identities.

Promotion still requires the independently frozen object/session experiment in #49 comparing framewise, within-window persistent, cross-window hypothesis, identity-marginalized, and exact-fallback arms. Joint-gauge residual evidence remains the separate #92 layer.


FlorianPfaff:
## Required dependency and semantic binding before merge

This stream is downstream of the pairwise result contract and should not merge from base `2d6df37...` ahead of PRs #92 and #94.

After those land, rebase and bind the interpreted pairwise semantics explicitly in every `MaterialIdentityAssociationSummaryV1` and update identity, not only indirectly through `association_result_id`:

- association descriptor schema name/version;
- base `ranking_semantics` (`isotropic-geometric-mutual-best-covariance-diagnostic-v1` after #92);
- whether covariance is an admission/ranking input or a diagnostic only;
- for joint-gauge evidence, the separate tracklet-producer revision, association revision, joint-gauge dependence semantics, and `conditional_point_cross_covariance_semantics`.

The result ID is necessary but not sufficient for portable interpretation: a consumer should be able to reject an unsupported ranking/dependence version without checking out historical source merely to discover what the digest meant.

Also add an adversarial migration test showing that a schema-v1/Mahalanobis-ranked pairwise result cannot be appended under the schema-v2 covariance-diagnostic stream contract, even when all endpoints and candidate numbers are syntactically valid. The stream should fail closed rather than silently combine hypotheses produced under incompatible ranking semantics.

Recommended merge order: #92 → #94 rebase/schema-v2 integration → #95 rebase and explicit semantic-version admission. This keeps append-only lineage from freezing the superseded covariance-volume ranking behavior.

FlorianPfaff:
A downstream consumer boundary is now proposed in #100.

PR #100 turns retained local alternatives into a source-calibrated discrete mixture with a mandatory newest-window null hypothesis, exact candidate-ID-aligned `logsumexp` likelihood marginalization, and Gaussian moment matching that retains between-identity covariance. This gives the append-only hypothesis stream a concrete non-rewriting consumer path.

Recommended integration order remains: land the hardened pairwise result contract first, then rebase this stream, then connect a frozen BayesianPhysTwin experiment to the mixture surface rather than promoting hard links to global identities.

PR #97, opened 2026-08-04
Route trusted CI recovery branches to workstation2

FlorianPfaff:
## What changed

- keep ordinary pushes, manual runs, and forked pull requests on `ubuntu-latest`;
- route only same-repository pull-request branches named `ci/self-hosted-*` to the existing `nvidia-smi` self-hosted runner label;
- apply the recovery route consistently to quality, Python 3.10–3.14, minimum-dependency, distribution-build, and installed-artifact jobs.

## Why

Run 30908119198 and its sibling workflows concluded without recording a single job step, and a retry remained queued without runner assignment. This is runner-capacity failure rather than a Python/test failure. The recovery branch convention provides a narrow, auditable escape hatch without exposing the self-hosted runner to forks or changing the normal GitHub-hosted default.

## Validation intent

This pull request itself uses the `ci/self-hosted-*` branch convention, so the complete existing test matrix should execute on workstation2 and provide fresh diagnostics for any real source failure hidden by the original runner outage.

### DIFF
--- a/.github/workflows/fail-closed-quality.yml
+++ b/.github/workflows/fail-closed-quality.yml
@@ -23,7 +23,12 @@ env:
 jobs:
   authoritative-quality:
     name: Authoritative Ruff and mypy
-    runs-on: ubuntu-latest
+    runs-on: >-
+      ${{ github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          startsWith(github.head_ref, 'ci/self-hosted-') &&
+          fromJSON('["self-hosted","Linux","X64","nvidia-smi"]') ||
+          'ubuntu-latest' }}
     timeout-minutes: 20
     steps:
       - name: Check out repository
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Recovery status:

- Original run 30908119198 and its sibling workflows completed/faulted without recording any job step or retained log blob, so there is no source-level failing command to patch.
- A failed-job retry of 30908119198 remains queued without runner assignment.
- This PR routes only same-repository `ci/self-hosted-*` pull requests to the established workstation2 labels `[self-hosted, Linux, X64, nvidia-smi]`; ordinary pushes, manual runs, and forks remain on `ubuntu-latest`.
- Current recovery runs: Tests 30946752991 and Fail-closed quality 30946753041. Both are accepted by Actions and waiting for workstation2.

Once a job is assigned, its fresh step/log data will distinguish any hidden source regression from the confirmed runner-start failure.

PR #110, opened 2026-08-05
Harden prediction storage and add joint evidence diagnostics

FlorianPfaff:
## Summary

This is a focused storage, dependence-calibration, evidence, and CI hardening batch. It does not change the frozen provider-v1/v2 observation contracts or promote a new scientific claim.

### Prediction-window storage

- version `PredictionWindow` NPZ archives and preserve validated dense precision by default;
- require an explicit choice for float32 compaction while retaining MotionCrafter's native float32 path;
- make schema-v2 archives closed and fail closed on unknown or missing members;
- require exact `int64` frame indices and Boolean masks for versioned archives;
- require `scene_flow` and `deform_mask` together;
- retain historical unversioned archive loading, including its float64 default and coercive legacy adapter.

### Joint dependence diagnostics

- evaluate `blockdiag(D) + U U.T` without constructing a dense `3N x 3N` covariance;
- replace the tall `3N x R` SVD with an `R x R` eigendecomposition of `U.T D^-1 U`;
- reuse the Gram Cholesky factor for the Woodbury correction;
- report joint NEES, NLL, log determinant, effective shared rank, and shared/conditional subspace energies with equal group weighting;
- read and evaluate the exact NPZ bytes whose SHA-256 is recorded;
- reject unknown NPZ members and publish reports with atomic create-if-absent semantics.

### Promotion evidence

- generate deeply validated, content-addressed JSON and Markdown evidence cards from held-out promotion locks and reports;
- reject ordinary tampering and re-signed semantic contradictions;
- compose evidence cards with deterministic held-out failure diagnosis and replay them through `heldout-provider verify`;
- publish both JSON and Markdown through the same atomic create-if-absent primitive so a concurrent writer is never overwritten.

### CI and compute evidence

- keep the authoritative quality workflow read-only and credential-free;
- isolate it from the self-hosted runner's global Python installation;
- pin the primary Ruff, mypy, pytest, pytest-cov, and NumPy versions used by that workflow;
- route the focused PR quality gate to `[self-hosted, Linux, X64, nvidia-smi]` for this recovery branch while retaining the isolated environment;
- retain a read-only compute-evidence workflow for large covariance, Monte Carlo, and full-resolution memory studies;
- remove all temporary write-enabled repair workflows before review;
- keep the quality policy aligned with the repository's established Ruff lint checks rather than introducing an unrelated whole-branch formatter migration.

- Frozen provider-v1/v2 schemas and estimator semantics are unchanged.
- Existing bundle loading retains its explicit in-memory dtype policy.
- Historical unversioned prediction archives remain readable.
- The joint-covariance command is a diagnostic, not a provider-promotion decision.
- Evidence cards summarize an existing frozen promotion report; they do not create a Causal4D intervention-benefit or state-of-the-art claim.
- Compute evidence is synthetic calibration/algorithmic stress or hardware-specific engineering evidence, not downstream physical-object benefit evidence.

## Validation

- 23 isolated regression checks pass for strict archives, scalable covariance diagnostics, and atomic evidence publication, including dense parity, direct-SVD subspace parity, no-tall-SVD enforcement, exact-byte evaluation, concurrent-writer preservation, legacy archive compatibility, malformed schema-v2 archives, and canonical non-overwriting JSON.
- Python compilation passes for the modified core modules and focused tests.
- The authoritative self-hosted run on source/test head `76f93cdf84172b45cfa53a48b47b366e5a5b262a` passed Ruff lint, mypy over 25 stable source files, and all `56/56` focused regressions. Its workflow conclusion was red only because a newly introduced `ruff format --check` requested formatting changes across unrelated existing branch files.
- Current head removes that accidental formatter-policy expansion without changing source or tests. The replacement quality run and the complete Python 3.10-3.14, NumPy-floor, source/wheel, installed-command, promotion, and compute-evidence workflows are queued.
- This PR has not been merged on the basis of partial or local checks alone.

Supersedes draft PR #109, whose GitHub-hosted runs remained queued.


FlorianPfaff:
The overlap with merged #108 is now reconciled: `heldout_promotion.py` and the documentation retain both deterministic diagnosis outputs and the new evidence-card outputs, and GitHub reports the branch mergeable.

Two fail-closed issues remain before this should leave draft:

1. `joint_covariance_metrics.run_joint_covariance_diagnostic` hashes `source.read_bytes()` but then reopens the path with `np.load(source)`. A concurrent replacement can therefore make `source_sha256` describe different bytes from the evaluated arrays. Load the NPZ from `io.BytesIO(payload)` instead.
2. Both new exclusive writers use a predictable `.<name>.tmp-<pid>` plus an existence check and `os.replace`. Concurrent writers in one process can delete each other's temporary file, and `os.replace` can overwrite a destination created after the check. Use a unique same-directory temporary file and an atomic no-replace publication primitive (for example `os.link(temp, destination)`), with a regression test that the concurrently created destination remains untouched.

Keep this draft until those are fixed and the exact-head CI matrix is green.

FlorianPfaff:
## Compute-heavy validation started

A revision-pinned GitHub Actions workflow now runs the expensive validation without cancelling when the PR receives later commits.

- **Compute evidence run:** `31032598440`
- **Hosted job:** large joint-covariance stress study, enlarged gauge-graph Monte Carlo, and finite-sample cycle-guard study
- **Self-hosted job:** full-resolution fusion and provider-evaluation memory profile on `[self-hosted, Linux, X64, nvidia-smi]`
- **Evidence:** checksummed JSON/CSV/log artifacts retained for 30 days

The run is bound to revision `0066bc119719ad5d7585b599db5d5f393c3fb0c2`. It remains scientifically scoped to synthetic calibration/algorithmic stress and hardware-specific engineering evidence; it does not establish downstream physical-object, BayesianPhysTwin, or Causal4D benefit.

FlorianPfaff:
## Final implementation and validation status

The requested correctness and scalability changes are now on the branch.

Current head: `db49a9f2bf7d13baa09641efdd0a1afb8c997add`.

Implemented in the final tranche:

- schema-v2 prediction archives now reject unknown/missing members, non-`int64` frame indices, non-Boolean masks, and unpaired flow/deformation masks while preserving the legacy coercive adapter;
- joint covariance diagnostics use the small `R x R` whitened Gram matrix rather than a tall `3N x R` SVD, reuse the Gram Cholesky factor, hash and evaluate identical NPZ bytes, and publish without clobber races;
- promotion evidence JSON and Markdown now share an atomic create-if-absent writer, with concurrent-writer and canonical-JSON regressions;
- the authoritative quality job is read-only, credential-free, isolated in a fresh virtual environment, and uses pinned quality-tool versions;
- temporary write-enabled repair workflows have been removed.

Authoritative run `31034767022` on the exact source/test head passed:

- Ruff lint: pass;
- mypy: 25 source files, pass;
- focused pytest: `56/56`, pass.

That workflow was marked failed only because an experimental `ruff format --check` attempted to impose formatting on unrelated existing branch files. The current head removes that accidental policy expansion without changing source or tests. Replacement quality run `31034924412` and the complete matrix remain queued. The PR is intentionally not merged while those current-head checks are outstanding.

PR #113, opened 2026-08-05
Harden observation-factor stream loading and append-only persistence

FlorianPfaff:
## Residual change

Earlier release-capsule work has already landed on `main`. The remaining four-file diff hardens the observation-factor stream boundary:

- strict JSON and schema validation for stream and referenced bundle manifests;
- exact-byte snapshots for bundle manifests and NPZ payloads before parsing;
- append-only persistence checks with an exclusive same-directory writer lock;
- no-overwrite first publication and a final byte check before replacement;
- adversarial regressions for duplicate keys, coercive values, pickle policy, append rollback/forks, and writer races.

## Compatibility

Valid schema-v1 stream IDs and update IDs remain unchanged.

## Validation note

The previous red capsule check was caused by this branch being behind `main`: a newer base workflow referenced a capsule test absent from the old head. The branch is being updated from `main` before final CI and merge.


FlorianPfaff:
Keeping this draft open because the strict-contract changes are valuable, but two fail-closed races remain before merge: (1) bundle manifests and NPZ payloads are hashed and then reopened for parsing, so a concurrent replacement can make retained hashes describe different bytes from those evaluated; parse a private exact-byte snapshot instead; (2) stream persistence validates the existing prefix before `os.replace` without serializing writers, so two concurrent appenders can both validate the same predecessor and lose/fork an update. Add same-directory writer serialization/CAS plus concurrent-writer regressions. Temporary repair workflows used during investigation have been removed from the branch.

PR #118, opened 2026-08-06
Add provider-neutral windowed 4-D source contracts

FlorianPfaff:
Superseded by merged PR #117, which adds the canonical provider-neutral prediction manifest, strict MotionCrafter adaptation, coherent visual-bias nuisance sidecars, source-calibrated common-mode stress evidence, CLI support, tests, and documentation. Keeping this alternative contract would create two competing provider-neutral surfaces, so this draft is closed without merge.


FlorianPfaff:
Closing as redundant rather than merging a third provider-neutral contract. This draft overlaps the same boundary as #117 and #119, remains red at its exact head, and introduces a separate `Windowed4DSourceManifestV1`/`prob4d.source` abstraction. The coherent-bias work in #117 is independently valuable, while the canonical provider contract should be consolidated into one API rather than three incompatible schemas.

PR #121, opened 2026-08-06
Add immutable artifacts and provider-neutral prediction bundles

FlorianPfaff:
## Summary

- replace bypassable frozen JSON subclasses with genuine read-only `Mapping`/`Sequence` values;
- move `PredictionWindow` arrays to immutable byte-backed storage and add exact-byte NPZ loading;
- add content-addressed provider-neutral prediction manifests/loaders, a MotionCrafter adapter, and grouped CLI routes;
- update provider-v2 and held-out consumers for the protocol boundary;
- add focused documentation, adversarial regressions, and a read-only CI lane.

## Compatibility

- no frozen provider-v1/v2, observation, factor-bundle, stream, or prediction archive schema/ID changes;
- valid JSON serialization and content identities are unchanged;
- in-memory frozen JSON values now implement `Mapping`/`Sequence`; callers needing mutable built-ins use `plain_json`;
- large execution-only `FusedSequence` and mmap ownership paths are intentionally unchanged.

## Security and causal boundaries

- exact source-manifest and member-byte binding;
- fail-closed digest, frame-grid, source-interval, symlink, traversal, and path-escape validation;
- causal source intervals are admitted before payload decoding;
- MotionCrafter’s original manifests, model-set identities, stochastic seed schedule, and temporal lineage remain authoritative.

## Validation

- all 686 source tests passed in bounded local groups;
- after final loader/CLI hardening, 85 directly affected tests passed;
- Python compilation, line-length, and whitespace checks passed;
- the added read-only workflow runs Ruff lint/format, mypy, and focused contract regressions on the exact reviewed head.

## Claim boundary

This is artifact-integrity, provider-boundary, causal-lineage, and interoperability infrastructure. It does not establish observation competence, uncertainty calibration, BayesianPhysTwin benefit, Causal4D benefit, safety, or state of the art.

## Scope coordination

This does not replace the append-only stream work in #113 or the release capsule in #114.

The current head contains a checksummed, self-removing publication capsule. A same-repository pull-request workflow verifies and rewrites it to one reviewable product commit before ordinary review and CI.


FlorianPfaff:
Closing as obsolete and non-reviewable. The live diff contains only six `.github/agent-patch/patch.b64.part*` archive fragments plus a workflow that reconstructs and applies them. The workflow fails before validation because the reconstructed archive no longer matches its pinned SHA-256. None of the intended implementation is present as a normal reviewable source diff, and subsequent provider-boundary work has moved on in `main` and PR #124.

PR #124, opened 2026-08-06
Add strict generic provider import

FlorianPfaff:
## Summary

Add a strict generic importer that materializes the existing canonical `PredictionProviderManifestV1` from external providers without adding their runtime stacks to Prob4D.

- accept versioned canonical `PredictionWindow` NPZ payloads plus a closed JSON import specification;
- require exact provider revision, provider-run, model-set, loader, dependence-group, and per-output causal-lineage semantics;
- derive payload hashes, byte counts, storage precision, and optional-field declarations from the admitted bytes;
- parse the specification and payloads from private exact-byte snapshots, then recheck source bytes before publication;
- reject traversal, symlinks, unknown or duplicate JSON fields, non-finite values, malformed identities, reserved metadata, and frame-lineage mismatches;
- retain the importer, example, tests, docs, and focused read-only workflow in source distributions.

## Usage

```bash
python -m prob4d.prediction_provider_import \
  outputs/sequence-a/provider-import.json \
  outputs/sequence-a/provider-neutral.json

prob4d prediction validate \
  outputs/sequence-a/provider-neutral.json \
  --causal-frame-stop 134
```

## Compatibility

- `PredictionProviderManifestV1` remains the single canonical portable prediction contract;
- MotionCrafter import, observation-provider artifacts, factor bundles, material identities, and frozen experiment decisions are unchanged;
- paths remain retrieval metadata; exact bytes and scientific/provenance semantics remain identity-bearing.

## Scope boundary

This PR does not alter the canonical manifest writer or verifier and does not claim multi-writer serialization. It is interoperability, provenance, causal-lineage, and admission hardening—not evidence of provider competence, calibration, independence, downstream benefit, deployment safety, or state of the art.


FlorianPfaff:
Superseded by #128. The original branch accumulated 38 repair commits and was repeatedly mutated by stale queued self-modifying workflows, reintroducing deleted temporary I/O scaffolding and merge conflicts. #128 was rebuilt from the current `main` with exactly the six reviewed files and no write-enabled repair automation.

PR #171, opened 2026-08-07
Add portable sparse gauge-tree artifacts

FlorianPfaff:
Superseded by PR #172, which is the authoritative portable sparse gauge-tree artifact implementation.

PR #172 retains the intended path-independent `O(K)` persistence and unchanged provider-v2/schema-v4 compatibility boundary, while adding stronger descriptor-bound no-follow reads, mutation detection, explicit manifest/member size caps, fixed portable dtypes, three-member factor storage, and installed-artifact/source-distribution validation.

The non-overlapping dense-memory status cleanup from this branch will be proposed separately. No code from this PR should be merged.


FlorianPfaff:
Closing as superseded by #172 after comparing the three concurrent portable-prior designs.

#172 is the authoritative path because it retains the sparse factors as three fixed-dtype, non-pickled NPY members and adds descriptor-bound no-follow reads, before/after file-identity checks, exact byte/shape/dtype/content identities, a 1 MiB manifest cap, and an exact-array-size plus 64 KiB header bound before allocation. It also includes installed-artifact and source-distribution validation.

The useful goals of this PR—path-independent sparse persistence, no dense covariance, no-clobber publication, strict replay, and unchanged provider-v2/schema-v4 bytes—are all retained there. Keeping two schema-v1 artifact names for the same prior would create avoidable downstream ambiguity. No source from this PR is being merged separately.

PR #173, opened 2026-08-07
Add portable sparse gauge-tree prior artifacts

FlorianPfaff:
## What changed

- add a strict, checksum-bound manifest/NPZ handoff for `GaugeTreeSquareRootPriorV1` without serializing the dense `7K × 7K` covariance
- preserve the existing path-independent `prior_id` and add a separate artifact identity over the prior, ordered gauges, parent tree, exact array descriptors, representation semantics, and optional source-covariance binding
- add immutable no-clobber writes and strict loading that rejects duplicate/non-finite JSON, coercive scalar aliases, path traversal, checksum changes, unexpected NPZ members, descriptor mismatches, malformed trees, and invalid Cholesky factors
- add `prob4d gauge prior verify` plus an explicit, size-guarded `materialize` compatibility path
- expose the artifact APIs publicly and register the grouped command
- document the contract and expand the focused workflow to run Ruff, formatting, MyPy, adjacent tests, and `compileall`

- no change to historical `GaugeTreeSquareRootPriorV1.prior_id` values
- no change to provider-v2 or schema-v4 `ObservationFactorBundle` identities
- schema-v4 factor bundles still carry the dense covariance; this PR supplies a standalone portable sparse-prior handoff without silently redefining that frozen contract
- the artifact verifies representation integrity and exact sparse algebra only; it does not establish provider competence, calibration, physical-query benefit, or deployment safety

## Validation

- `86 passed` in the focused and adjacent local suite covering gauge-tree algebra, observation-space actions, artifact I/O, CLI/registry routing, factor bundles, and sparse observation factors
- `git diff --check` and `python -m compileall -q src/prob4d tests` passed
- Ruff and MyPy are included in the pull-request workflow

### DIFF
--- a/.github/workflows/sparse-gauge-tree-prior.yml
+++ b/.github/workflows/sparse-gauge-tree-prior.yml
@@ -9,7 +9,12 @@ on:
       - "docs/sparse-observation-factor-stack.md"
       - "src/prob4d/_gauge_tree_*.py"
       - "src/prob4d/gauge_tree_prior.py"
+      - "src/prob4d/gauge_tree_prior_io.py"
+      - "src/prob4d/command_registry.py"
+      - "src/prob4d/cli.py"
       - "tests/test_gauge_tree_*.py"
+      - "tests/test_cli.py"
+      - "tests/test_command_registry.py"
       - "tests/test_sparse_observation_factors.py"
   workflow_dispatch:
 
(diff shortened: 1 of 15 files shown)


FlorianPfaff:
Closing as superseded by #172 after comparing the concurrent sparse-prior handoff contracts.

The grouped CLI and public routing here are useful, but the single NPZ payload introduces a weaker intake boundary than #172's bounded standalone NPY members: the payload is checksum-verified and then reopened for decoding, while compressed-container member expansion is not bounded by the manifest's exact per-array byte count before NumPy opens it. #172 instead reads each fixed-dtype member once through a no-follow descriptor, bounds it to exact array bytes plus a small NPY-header allowance, verifies before/after file identity, and decodes from the verified in-memory snapshot.

#172 will be the single schema-v1 portable prior format. A grouped `prob4d` command can be added after that artifact API merges, without retaining a second incompatible persistence schema. No target or frozen provider behavior is affected.

PR #233, opened 2026-08-14
Add recurrent-online CUT3R provider and harden public surfaces

FlorianPfaff:
## Summary

- add a runtime-neutral importer for official recurrent-online CUT3R depth, confidence, and camera outputs;
- publish one content-addressed provider-neutral `PredictionWindow` with prefix-growing causal lineage;
- bind exact CUT3R revision, checkpoint, input-video, adapter, and source-member identities;
- expose the importer through `prob4d prediction import-cut3r-online`;
- add focused malformed-input, geometry, no-clobber, provenance, and causal-admission tests;
- add randomized `Sim(3)` group-action and covariance-preservation regression tests;
- validate active documentation against the grouped command registry and generated `prob4d.api.v2` manifest; and
- migrate active examples away from removed Prob4D 0.4 executables and package-root imports.

## Validation

Exact-head GitHub Actions passed on revision `909579174ddb77b0465988ce64ffa1db722c8340`:

- complete NumPy-core suites on Python 3.10, 3.11, 3.12, 3.13, and 3.14;
- declared minimum runtime dependencies on Python 3.10;
- repository, release, API, documentation-surface, Ruff, and MyPy checks;
- wheel and source-distribution build validation plus isolated installed-artifact CLI smoke tests;
- focused provider-runtime loading, coordinate-dispatch, conservative-fusion, and clean-tree checks;
- generic-provider import, provider admission/evaluation, fail-closed quality, and security scanning workflows.

The Python 3.12 complete-suite run reported 1,381 passed and 7 optional-`h5py` skips. The official CUT3R `demo.py` exporter was also checked to confirm the imported `depth/%06d.npy`, `conf/%06d.npy`, and `camera/%06d.npz` layout with `pose` and `intrinsics` members.

## Scientific boundary

This change establishes provider interoperability, exact provenance, causal-lineage validation, and regression coverage only. It does not establish CUT3R or Prob4D provider competence, metric scale, covariance calibration, BayesianPhysTwin benefit, Causal4D intervention benefit, deployment safety, or state of the art. The existing support, source-mean, identity, gauge/dependence, point-covariance, query-relevance, target-admission, and exact-fallback gates remain unchanged.

PR #234, opened 2026-08-14
Add provider adapter SDK and prospective readiness matrix

FlorianPfaff:
## Summary

- add a typed `PredictionProviderAdapterV1` SDK that materializes external provider outputs into the canonical `PredictionProviderManifestV1` boundary;
- add deterministic adapter conformance checks for repeatability, ordering, causal-prefix invariance, exact identity binding, payload mutation, and declared dependence semantics;
- add a prospective multi-provider readiness matrix that reuses the existing seven ordered readiness gates and issues one target authorization per provider only after a complete pass;
- expose both additions through canonical grouped `prob4d prediction ...` routes;
- add documentation, examples, focused tests, and a dedicated exact-head workflow; and
- keep Prob4D independent of BayesianPhysTwin and Causal4D at runtime while binding their query/fallback identities in the frozen readiness protocol.

## Why

Prob4D already has strong provider-v2 contracts and failure-localization machinery, but adding a backend still required bespoke manifest construction, and comparing several prospective providers required hand-assembled readiness artifacts. This PR turns those two repeated integration tasks into reusable, content-addressed infrastructure without adding another estimator or opening protected target outcomes.

The readiness matrix keeps the existing stop order:

1. support feasibility;
2. source mean;
3. identity/reliability;
4. gauge/dependence;
5. conditional point covariance;
6. downstream query relevance; and
7. exact fallback.

A provider that fails one stage terminates there. Later gates remain unevaluated, and no target authorization is emitted.

## Validation

The branch includes focused unit and CLI coverage for adapter materialization, conformance failure modes, readiness ordering, provider-order invariance, one-shot authorization, and strict replay. The dedicated workflow also runs Ruff, Ruff formatting, MyPy over the new typed boundaries, focused regressions, installed command smoke tests, and the existing CLI suite.

The first workflow execution failed in Ruff before MyPy or tests. Canonical Ruff fixes and formatting were then applied on the PR branch. Temporary branch-restricted formatter/validation helpers were removed again. The latest exact-head workflow results and the finalizer's machine-readable PR comment are the source of truth for review readiness; a failed, cancelled, queued, or timed-out workflow does not authorize the PR to leave draft state.

## Scientific boundary

This change is interoperability and prospective orchestration infrastructure. It does not establish provider competence, calibrated uncertainty, BayesianPhysTwin physical benefit, Causal4D intervention benefit, deployment safety, or state of the art. A future provider study must still freeze fresh object/session units, source/calibration policies, downstream query and fallback identities, and open its bound target roster only after the complete readiness sequence passes.


FlorianPfaff:
Closing this draft as superseded by the concrete provider-integration path in #233.

This PR introduces a large second abstraction layer (adapter SDK, conformance framework, readiness matrix, new versioned public surface, and dedicated workflow) before we have evidence from multiple real providers that the existing provider-v2 boundary creates repeated integration pain. It also never obtained executable CI evidence: the current exact-head runs remained `action_required` rather than exercising the implementation.

#233 instead advances the immediate scientific and engineering priority: one real recurrent-online CUT3R importer with exact provenance, causal lineage, malformed-input coverage, and the existing ordered readiness gates. Keeping both branches would duplicate CLI/public-surface work and increase maintenance risk.

The generic SDK idea can be reconsidered after at least two independently integrated providers expose the same concrete boilerplate. At that point, the reusable subset should be extracted from working integrations rather than merged prospectively.

Issue #204, opened 2026-08-11
Localize whether a richer point-uncertainty model is scientifically authorized


FlorianPfaff:
Use only already-open source/diagnostic evidence to decide whether point-uncertainty v2 development is justified.

The localization must distinguish:

1. unsupported geometry or incomplete provider support;
2. observation-mean or association failure;
3. gauge/dependence misspecification;
4. conditional point-covariance miscalibration after useful means and identities are established; and
5. downstream query non-identifiability.

A richer point-covariance model is authorized only for case 4, using source/calibration groups only. Cases 1, 2, 3, or 5 stop or redirect the provider version. The opened 19-interaction MotionCrafter cohort and opened Deform360 calibration objects remain diagnostic and cannot become confirmation.

FlorianPfaff:
The software localization prerequisite is now implemented in #211 and locked by the conformance corpus in #212. `point-covariance-localized` is the only terminal classification that authorizes source/calibration-only point-uncertainty development; support, source-mean, identity/association, gauge/dependence, and query-relevance failures do not. This issue remains open because authorization still requires the actual already-open source/diagnostic evidence to produce that classification; no richer model has been developed or selected from target outcomes.

FlorianPfaff:
Implementation update: PR #222 was repaired, validated, and squash-merged as `0dc29984caa0924235357a7b0cf1a2e081134cae`.

The merged code adds the missing fail-closed, content-addressed **disjoint source-validation promotion gate** for `PointUncertaintyCalibrationV2`. It compares one frozen v2 calibration against the matched production-v1 calibration on complete independent validation groups, replays proper score, 90% coverage, covariance width, normalized energy, group wins, and worst-group harm, and rejects training/validation overlap or artifact drift.

This does not enable v2 in provider-v2 exports and does not resolve this issue by itself. The localization rule remains unchanged: v2 is scientifically eligible only for conditional point-covariance miscalibration after useful means, identities, support, and gauge propagation are established. No target or confirmation outcomes were accessed.

FlorianPfaff:
Decision: **a richer conditional point-covariance model is not scientifically authorized for the completed provider versions.**

[...]

PR #242, opened 2026-08-17
[Superseded by #246] Add structured covariance queries and pathwise diagnostics

FlorianPfaff:
## Superseded

Closed in favor of #246, which preserves the evidence-safe structured covariance-query work on a fresh branch from current `main` and removes the scientifically unsafe track-level calibration surface.

The original first-axis trajectory/track design treated multiple tracks from one physical object or acquisition session as independent conformal units. That would pseudo-replicate shared provider, gauge, scene, action, and calibration errors and could make finite-sample ranks and simultaneous-coverage reports overconfident.

#246 therefore contains only:

- exact matrix-free covariance actions for dense, sparse, and tree-sparse stacks;
- exact `A @ Sigma @ A.T` query projection with conditional/shared-gauge decomposition;
- scalar quadratic forms;
- stable `prob4d.api.v2` exports;
- strengthened regression tests and explicit ownership/evidence documentation.

Any future pathwise calibration must be designed separately around immutable complete-object or acquisition-session groups, with group-level fitting and evaluation. Tracks, points, frames, views, and taxels must not be counted as independent replicates.


FlorianPfaff:
## Scientific-unit blocker before merge

The structured covariance-action and `A @ Sigma @ A.T` projection work is useful and appears aligned with Prob4D's explicit-gauge contracts. I am not merging the current pathwise-calibration surface as written, because it defines each first-axis **trajectory/track** as one exchangeability unit and calibrates the split-conformal threshold over the resulting trajectory maxima.

That conflicts with the frozen Prob4D/BayesianPhysTwin evidence boundary used by #49 and #204: independent calibration and evaluation units must be complete physical objects or acquisition sessions, never tracks, points, frames, views, or taxels. Multiple tracks from one object/session generally share provider, gauge, scene, action, and calibration errors; counting them independently can make the finite-sample rank and reported simultaneous coverage materially overconfident.

A safe resolution is one of:

1. **Split the PR:** retain the matrix-free covariance queries, but remove the pathwise calibration and its `prob4d.api.v2` exports until the grouped design is ready; or
2. **Make calibration group-aware:** require immutable `group_ids` (object/session IDs), aggregate the maximum statistic across every admitted trajectory and valid step inside each complete group, compute the conformal rank from the number of independent groups, and report target coverage/equal weighting by group. The artifact must bind the exact group roster and forbid track-level pseudo-replication.

The target diagnostic may still report per-trajectory failure runs descriptively, but any finite-sample simultaneous-coverage claim must be fitted and evaluated at the frozen independent-group level. Because `prob4d.api.v2` is a stable ecosystem façade, this should be corrected before adding the symbols rather than repaired later with a silent semantic change.

The branch is also behind current `main` (`e2f196174174b6f11481610e131d24ecdbea1ebf`) and should be refreshed only after this semantic decision. No target data is needed to make the correction.

PR #252, opened 2026-08-18
Add provider-neutral held-out promotion locks

FlorianPfaff:
## Summary

- add an additive provider-neutral v2 held-out promotion lock aligned with `PredictionProviderManifestV1`
- preserve historical MotionCrafter v1 lock descriptors, content IDs, reports, and evidence cards through the exact original implementation blob
- require complete provider-contract agreement for v2 target admission
- emit provider-neutral v2 promotion evidence cards
- add focused compatibility, drift-rejection, documentation, and example coverage

## Compatibility

Historical v1 artifacts remain readable and are never rewritten. V2 replaces only the top-level `motioncrafter_revision` and `model_set_id` fields with a nested, versioned provider identity. Existing evaluator, guarded-query, bootstrap, and exact-fallback semantics are unchanged.

The version dispatcher retains the exact pre-change evidence-card implementation as `_promotion_evidence_v1.py`; v1 callers and content identities continue through that unchanged logic, while v2 cards replace only the provider-specific repository record.

## Validation

- local Python compilation passed for all changed source and tests
- the provider-neutral example JSON parses successfully
- every uploaded Git blob matched the locally validated Git object hash
- repository Actions are the authoritative Ruff, mypy, pytest, package, provider, ecosystem, portability, and security validation

## Scientific and information boundary

This changes provider/lock identity and target-admission validation only. It opens no source, calibration, target, or held-out outcome; changes no estimator, calibration, comparison arm, guard, fallback, or scientific claim; and produces no physical evidence.

### DIFF
--- a/CHANGELOG.d/provider-neutral-promotion-lock.md
+++ b/CHANGELOG.d/provider-neutral-promotion-lock.md
@@ -1,0 +1,7 @@
+Add an additive provider-neutral version of the held-out promotion lock and its
+paper-facing evidence card. New locks freeze the complete
+`PredictionProviderManifestV1` contract for CUT3R, VGGT, MotionCrafter, or future
+providers, and target admission checks every invariant provider field. Historical
+MotionCrafter lock and evidence-card descriptors remain byte- and identity-stable.
+This changes no estimator, calibration, target access, fallback, or scientific
+claim.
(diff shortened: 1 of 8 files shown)


FlorianPfaff:
Superseded by main commit `59588d1f7930df19eeee98c6b7e8b368d4df4b41`, which independently implements the provider-neutral promotion-lock, target-admission, evidence-card, documentation, and regression path. Closing this duplicate rather than competing with the merged implementation. A separate narrow hardening change will cover the remaining coercive nested `schema_version` case.

Issue #333, opened 2026-08-27
Qualify PointWorld on Flat'n'Fold before target access


FlorianPfaff:
## Objective

Execute the source-only qualification frozen in `protocols/pointworld-flatnfold-source-qualification-v1.json` before any Flat'n'Fold target outcome is opened.

This is the proposed fresh-provider path for issue #49: PointWorld -> Prob4D -> BayesianPhysTwin on a later garment-disjoint Flat'n'Fold robot cohort.

## Frozen source revisions

- Prob4D: `d02f057671023ff586ebfc15c904dbb0a60f4425`
- BayesianPhysTwin: `b5f07d649ac2cd7dc6ca1aceb4004ff4803bddc8`
- PointWorld: `05484826dfef74cbe278a3974179a5a16705d35d`
- Flat'n'Fold code: `fa0d3d17ac827e7b5c43ec1ef7c0c38ad5e39340`

## Representation gate — software decision completed in PR #334

PointWorld predicts seeded scene-point trajectories/flows on `B x T x N_scene x 3` with a `B x T x N_scene x 1` log-variance output. The existing dense `PredictionWindow` v2 contract is not used.

PR #334 implements and freezes `sparse-persistent-point-contract-v1`:

- persistent integer point identities and seed positions;
- seed-relative displacement trajectories;
- separate visual-source and planned-action lineage;
- explicit dependence groups;
- raw PointWorld log variance retained without a metric-covariance claim;
- no dense rasterization, target-truth nearest-neighbour mapping, or target-tuned interpolation.

The machine-readable decision is `protocols/pointworld-sparse-representation-v1.json`; implementation and tests are under `pointworld_sparse_prediction.py`, `pointworld_sparse_adapter.py`, and `test_pointworld_sparse_prediction.py`.

## Source-only checklist

- [ ] Bind exact PointWorld checkpoint SHA-256 and runtime environment.
- [ ] Bind exact PointWorld normalization-statistics SHA-256.
- [ ] Bind exact Flat'n'Fold dataset-byte manifest for the inventory/source cohort.
- [ ] Freeze complete garment identities before provider residuals.
- [ ] Verify three-camera timestamps, intrinsics, and extrinsics.
- [ ] Verify Baxter action-pose semantics and construct PointWorld robot point-flow actions without target state.
- [ ] Verify metric coordinate transformations using released calibration only.
[...]

PR #337, opened 2026-08-27
Add persistent-point provider artifact for PointWorld

FlorianPfaff:
## Purpose

Continue issue #333 by resolving the PointWorld representation mismatch without target-dependent rasterization.

The released PointWorld model keeps one seeded `N_scene` point axis over a forecast window. This PR adds a strict versioned artifact that preserves those identities and absolute predicted positions directly.

## What is added

- `src/prob4d/persistent_point_prediction.py`
  - immutable `PersistentPointPredictionWindow`;
  - strict versioned NPZ round trip;
  - deterministic window-scoped point IDs that do not assert cross-window identity;
  - PointWorld conversion from `scene_flows`, validity, and scalar `log_var` arrays;
  - explicit storage of provider uncertainty semantics and normalization-statistics SHA-256.
- `tests/test_persistent_point_prediction.py`
  - round-trip, identity, dtype, finite-value, malformed uncertainty, source-index, and unknown-field tests.
- `docs/persistent-point-prediction-window.md`
  - representation, identity, uncertainty-calibration, and claim boundaries.
- `evidence/pointworld-flatnfold-source-qualification-v1/representation-decision.json`
  - content-addressed source-only decision; no prediction payload, residual, target, BayesianPhysTwin, or Causal4D outcome was opened.
- focused GitHub Actions validation.

## Scientific decision

The representation path is now **artifact-authorized but the provider is not yet qualified**.

PointWorld constructs absolute positions as its initial scene points plus predicted normalized-relative displacement. The same `N_scene` axis persists over the output horizon. Its scalar log variance is trained in normalized relative-displacement coordinates, so this PR deliberately retains it as raw provider evidence and does not call it metric covariance.

No dense `H x W` adapter is introduced. Points from different forecast windows receive different identities until an explicit cross-window association is established.

## Remaining gates

This PR does not bind the exact checkpoint, runtime, normalization bytes, Flat'n'Fold dataset manifest/garment roster, camera/action transforms, visual/action lineage, uncertainty calibration, or `ProviderSupportFeasibilityV1` result. Those remain source-only work under #333 before any target outcome.

## Validation performed before push

A standalone execution of the new module and test file passed `7/7` focused tests and Python compilation. The PR workflow adds repository lint, formatter, mypy, evidence-digest, observation-factor compatibility, and compile checks.

## Claim boundary

This is interoperability and source-representation evidence only. It does not establish PointWorld competence, calibrated uncertainty, Prob4D fusion benefit, BayesianPhysTwin benefit, Causal4D benefit, or deployment safety.

### DIFF
--- a/.github/workflows/persistent-point-prediction.yml
+++ b/.github/workflows/persistent-point-prediction.yml
@@ -1,0 +1,96 @@
+name: Persistent-point prediction contract
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/persistent-point-prediction.yml"
+      - "docs/persistent-point-prediction-window.md"
+      - "evidence/pointworld-flatnfold-source-qualification-v1/representation-decision.json"
+      - "src/prob4d/persistent_point_prediction.py"
+      - "tests/test_persistent_point_prediction.py"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: persistent-point-prediction-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  PIP_DISABLE_PIP_VERSION_CHECK: "1"
+  PYTHONUNBUFFERED: "1"
+
+jobs:
+  contract:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Check out exact reviewed source
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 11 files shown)


FlorianPfaff:
Closing as superseded/scope-mixed after PR #336 merged the persistent PointWorld contract. This branch is now substantially behind `main`, conflicts with the merged persistent-point implementation, and also bundles a separate dual-gripper surface-action implementation. The persistent-point portion is redundant; if the dual-gripper action remains useful, it should be resubmitted as a focused PR against current `main` so it can be reviewed and validated independently.

PR #338, opened 2026-08-27
Add query-conditioned observability for partial 4D gauges

FlorianPfaff:
## Purpose

Turn the recently added observable-subspace Sim(3) factor into a decision-relevant scientific kernel.

A rank-deficient overlap may locate points on its observed support while leaving an off-support physical query dependent on an unobserved gauge direction. Pooled point error and a favorable average covariance reduction do not identify that distinction. This PR projects a partial gauge likelihood through a declared downstream query and makes exact fallback possible before BayesianPhysTwin consumes the candidate.

This is the method-level bridge needed by #49 and the PointWorld/Flat'n'Fold path in #333. It does not open any protected target outcome.

## Method

For a query Jacobian `J_q` in the centroid-normalized Sim(3) chart and a declared positive-definite output metric, the new kernel reports:

- the fraction of query sensitivity lying directly in the factor's observable subspace;
- the complementary sensitivity to the retained geometric nullspace;
- complete-prior query covariance before and after factor fusion;
- metric-weighted query-variance reduction; and
- the worst posterior/prior variance ratio over the supported query covariance.

Direct geometric support is deliberately separated from variance reduction transmitted only through prior correlation. A source-frozen three-criterion gate returns stable rejection reasons but does not alter the factor, invent a ridge, or implement fallback; BayesianPhysTwin remains responsible for exact caller-owned fallback.

A point-position Jacobian helper covers endpoint, marker, contact, and probe-position queries. Arbitrary nonlinear physical queries can supply their own local Jacobian and output metric.

## Deterministic failure control

The checked evidence isolates one unobservable rotation-about-line direction under a complete identity prior and one frozen gate:

| Case | Rank | Direct support | Variance reduction | Worst ratio | Decision |
|---|---:|---:|---:|---:|---|
| Point on observed line | 6 | 1.000 | 0.909 | 0.091 | admit |
| Distant off-axis probe | 6 | 0.679 | 0.618 | 0.965 | reject |
| Invalid full-rank completion, same probe | 7 | 1.000 | 0.909 | 0.091 | admit |

The last arm is intentionally invalid. It demonstrates that query conditioning cannot repair a factor that has already fabricated information in its nullspace; preserving rank deficiency is necessary.

## Files

- `src/prob4d/query_observability.py`: reports, gate, and point-query Jacobian;
- `src/prob4d/query_observability_study.py`: deterministic analytic control and CLI;
- `evidence/query-observability-control-v1/result.json`: checked mechanism evidence;
[...]

PR #341, opened 2026-08-29
[Superseded by #343 and #347] Audit finite gauge ambiguity beyond local query observability

FlorianPfaff:
## Scientific question

Can a locally supported physical query still vary over a finite unobserved gauge, and can decision-relevant common uncertainty cancel even when marginal intervals are broad?

This follows the partial-gauge and local-query work in #332 and #338. It is an **analysis-only method extension and deterministic development control**, not another provider adapter or a change to a target-side admission policy. It does not claim to repair a localized real-provider failure or advance a provider readiness stage.

## Concrete counterexample

An exactly collinear x-axis overlap is invariant under rotation about x. For a probe initially at `(0, 0.1, 0)`, the scalar query is `y(phi) = 0.1 cos(phi)`. Its first twist derivative at zero vanishes, while its finite range is `[-100, 100] mm`.

The existing rank-six analytic factor gives direct local support 1, variance reduction 10/11, and worst variance ratio 1/11; the existing illustrative local gate admits the scalar query. This is **not a bug in that explicitly local calculation** and does not establish posterior undercoverage. It falsifies interpreting the local report as global finite-orbit invariance or a distribution-free decision certificate.

## Method

Adds a NumPy-only experimental kernel for registered axial rotation groups and affine queries:

- exact finite-orbit query and pairwise-contrast bounds;
- correct aggregation inside a shared gauge group before taking amplitudes;
- sharp worst-case regret for actual affine action losses sharing the same angles;
- an attaining competitor/angle witness; and
- explicit support-motion and malformed-input checks.

The action check is analysis only: no action, belief selection, provider export, or production admission rule is changed. BayesianPhysTwin retains complete-belief selection and exact fallback ownership.

## Controls

| Control | Analytic result |
|---|---|
| Stationary off-axis y query | local twist derivative 0; finite range `[-100, 100] mm` |
| Off-axis probe, axial-coordinate query | exactly `40 mm` for every angle |
| Equal-radius y difference, shared angle | exactly 0 |
| Same points, separately variable angles | `[-200, 200] mm`; coincident axes do not authorize merging |
| Losses `0.1-y` and `0.1+y` | reference-selected action has `200 mm` worst-case regret; exceeds illustrative `50 mm` budget |
| Losses `0.05+y` and `0.10+y` | overlapping marginal intervals, but action 0 has exactly zero regret |

## Assumptions and novelty boundary

Sharpness is restricted to the declared product of full axial circles, fixed remaining transform parameters, and affine queries/losses. A rank-six factor, a near-collinear support, or a small Jacobian does not establish a finite likelihood symmetry. Smaller/coupled domains give conservative rather than necessarily sharp bounds. This is float64 evaluation of analytic formulas, not certified interval arithmetic.

Rodrigues' formula, harmonic support functions, group symmetry, and worst-case regret are classical. The documentation cites X-ICP and symmetry-reduced filtering and does not claim an exhaustive novelty result. The proposed contribution is the narrowly scoped finite-gauge/shared-lineage audit and its concrete negative and positive controls, not a new generic quotient theorem or a fourth manuscript.

## Compatibility and evidence boundary

Exactly four additive files: kernel, reproducer, tests, and derivation/usage note. No stable API, existing result, calibration, exporter, workflow, protected dataset, source/target roster, runtime request, or `claims.json` change. No provider, target, BayesianPhysTwin, or Causal4D experiment executed. No new hardware acquisition is assumed. #333 remains separately governed and unqualified.

## Validation before publication

- 27 standalone tests passed locally on Python 3.13.5 / NumPy 2.3.5.
- One existing-module integration test was explicitly deselected in the minimal offline copy; the committed test does **not** silently skip and must pass in the full repository CI.
- Independent quaternion reference, attained extrema/regret witnesses, shared cancellation, frame and axis-sign invariance, no-free-angle case, invalid inputs, and deterministic development output checked.
- Python compilation passed.
- Local Ruff/mypy were unavailable; exact-head hosted CI remains authoritative.

Draft pending exact-head CI and inspection. No auto-merge requested. Paper-facing development output will be retained separately without promoting an empirical claim.

### DIFF
--- a/docs/finite-gauge-query-ambiguity.md
+++ b/docs/finite-gauge-query-ambiguity.md
@@ -1,0 +1,210 @@
+# Finite axial-gauge ambiguity for physical queries
+
+## Purpose and status
+
+This experimental, NumPy-only analysis extends the **local**
+[query-observability diagnostic](query-conditioned-observability.md) with exact
+finite-orbit bounds for a restricted but useful geometry: one free rotation
+about a registered line per shared gauge group, with all other transformation
+parameters fixed. It also supplies sharp worst-case regret for actual affine
+action losses that share those same gauges.
+
+The existing local diagnostic is not mathematically incorrect. The new control
+shows why its direct-observability fraction must not be interpreted as a global
+invariance or decision certificate. A small first derivative does not establish
+that a query is constant on the finite observation-equivalence class.
+
+This is analytic development evidence, not real-provider competence,
+calibration, BayesianPhysTwin benefit, Causal4D benefit, or deployment safety.
+No provider, dataset, protected target, or physical execution is used. No
+existing result, production exporter, stable API, or admission rule is changed.
+
+## A falsifying control with a positive counterpart
+
+An exactly collinear overlap lies on the x-axis. Rotation about that axis leaves
+all its point coordinates unchanged. A probe initially at `(0, 0.1, 0)` has
+
+$$q(\phi)=0.1\cos\phi,\qquad q'(0)=0,\qquad q''(0)=-0.1.$$
+
+For the existing analytic rank-six factor, the scalar y-position Jacobian has
+zero projection onto the missing rotation-x direction. With its complete
+identity prior and observable precision 10, the local report gives direct
+support 1, variance reduction 10/11, and worst variance ratio 1/11. The existing
+illustrative `(0.8, 0.8, 0.5)` gate admits this scalar query. Nevertheless, the
+finite y-query range over a full rotation is `[-100, 100] mm`.
+
+These comparisons concern **different guarantees**. A full-circle ambiguity
+bound is deliberately stronger than a prior-weighted local uncertainty
+calculation. The control does not establish that the local posterior's actual
+credible interval undercovers or that its Bayesian decision is wrong.
+
+There are also useful positive controls. The axial coordinate of an off-axis
+probe remains fixed. Two points with the same transverse displacement have an
+exactly invariant y-difference when they share one angle. Treating their angles
+as separately variable destroys that cancellation; falsely merging unrelated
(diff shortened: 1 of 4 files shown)


FlorianPfaff (commented):
Validation completed for head `9f3cab5db1dcb09841ce53d448b9ce657379be69`.

- Full hosted matrix passed: Python 3.10-3.14 plus declared runtime floors; build/install smoke jobs also passed.
- The Python 3.12 job `99171359127` in [Tests run 33279210423](https://github.com/IPS-Stuttgart/Prob4D/actions/runs/33279210423) reports **1908 passed, 10 existing skips, 2 existing warnings**. `tests/test_axial_gauge_query.py` executed all **28 tests**, including the real local-gate integration; none of those tests were skipped.
- Fail-closed quality run `33279210438`, provider preflight `33279210432`, and security scanning `33279210425` all passed. The quality workflow checks changed Python with Ruff and the existing stable-interface list with mypy; this is not a claim of full new-module mypy coverage.
- CI checked out synthetic merge `19d4a2e70c17d2604f04d4c8296fd93eb0ac22ab`. GitHub reports tree `dbf050a408e3ad8603500d7b426aeb024138370c`, identical to the PR head's tree.
- All four pushed files' Git-blob identities match the locally tested files.

Paper-side development output, source/output hashes, scope, and CI provenance are retained in [BayesianPhysTwin-Paper PR #115](https://github.com/FlorianPfaff/BayesianPhysTwin-Paper/pull/115). This supersedes the initial pending-CI paragraph, without promoting an empirical claim.

The attempted ready-for-review transition hit a connector GraphQL schema error (`Repository.fullDatabaseId`). No merge or auto-merge was requested.

Issue #339, opened 2026-08-27
Promote query-aware observability from analytic control to held-out evidence


FlorianPfaff:
## Scientific objective

Build the evidence required for a Prob4D-centered paper claim:

> Partial gauge information is admitted only when it supports a declared physical query; complete-prior uncertainty is retained in unobserved directions; unsupported precision is rejected; otherwise the exact physical fallback is preserved.

The implementation is on `science/query-aware-observability-v1`. This issue tracks promotion beyond controlled mechanism evidence.

## Stage 1 — implementation and analytic control

- [ ] Query-aware implementation PR is green under focused and repository CI.
- [ ] Analytic collinear control artifact is frozen with source revision.
- [ ] On-support query is `factor-supported`.
- [ ] Broad-prior off-axis query is `fallback-unresolved`.
- [ ] Tight-prior off-axis query is `prior-bounded`.
- [ ] Bogus full-rank completion is `fallback-invalid-update` despite sub-tolerance reported uncertainty.

## Stage 2 — DLO-like controlled sweep

Freeze before running:

- [ ] effective rank and geometry-thickness grid;
- [ ] overlap/noise/prior-strength grid;
- [ ] on-support, near-support, and off-support query families;
- [ ] component-wise query tolerances;
- [ ] rank, uncertainty, gain, information-mismatch, and unsupported-information thresholds;
- [ ] sequence/trial seeds and artifact schema.

Required baselines:

- [ ] reject all rank-deficient factors;
- [ ] point transform with no covariance;
- [ ] pseudoinverse covariance without query admission;
- [ ] ridge/full-rank completion;
- [ ] exact observable-subspace fusion without query admission;
- [ ] query admission without provenance audit;
- [ ] full proposed method.

Required outputs:

- [ ] query RMSE/NLL/coverage/calibration;
- [ ] false-admission and false-rejection rates;
- [ ] fallback and harmful-update rates;
- [ ] risk/coverage curves;
- [ ] results conditional on rank and query-nullspace sensitivity;
- [ ] runtime and route accounting.

## Stage 3 — fresh-provider qualification

Coordinate with #333 and #49.

- [ ] Reconcile the overlapping PointWorld persistent-point draft PRs into one minimal source contract; do not merge duplicates.
- [ ] Freeze provider checkpoint/runtime/dataset hashes.
- [ ] Pass camera synchronization and calibration gates.
- [ ] Pass metric-coordinate and persistent point-identity gates.
[...]

PR #345, opened 2026-08-29
Audit complete Gaussian query updates and test the partial-observability trade-off

FlorianPfaff:
## Scientific purpose

Advance the controlled-method stage of #339 without duplicating the existing query-conditioned observability kernel or opening any provider/data cohort.

The existing query gate can be fooled by a fabricated full-rank factor. There is a second failure mode: a proposal can have the correct covariance while changing an unsupported mean. This additive experimental module checks BOTH Gaussian natural-parameter equations against an independently retained likelihood, then distinguishes factor-supported, prior-bounded, unresolved, and invalid-update routes.

## Implemented

- Prior-whitened, origin-centered covariance AND natural-parameter audit; no covariance-only loophole.
- Absolute joint query-precision gate, separate geometric support and prior-bounded routing, and explicit no-calibration/no-harm-guarantee boundaries.
- Adversarial tests, independent measurement-space Kalman comparison, coordinate/origin invariance, and an integration test for the existing ObservableGaugeFactor.
- A CPU-only local linear-Gaussian study with DLO-like point geometry, 48 configurations, 128 independent paired seeds, three queries, four proposal conditions, two prior worlds, and eight comparator arms.
- Raw sufficient statistics, proposals, decisions, complete-seed scores, width/coverage/proper-score metrics, and seed-clustered paired bootstrap intervals.

## Registration and execution

Protocol and exact implementation hashes were committed at `eb846570085cb33ceb7cfcfc9ae7cbcac01fadbe` BEFORE the first simulation execution. The study completed locally with those hashes; an independent raw-array verification and paper-facing interpretation are being added in this PR. No numerical setting has been changed after viewing outcomes.

The early result is deliberately bounded: correct partial factors enable useful query admissions; fabricated information is rejected; fallback has substantial point-error costs; and a wrong prior remains undercalibrated. This is NOT a claim to beat the correctly specified ungated Gaussian posterior.

## Validation boundary

Local focused tests: 17 passed, 1 integration test skipped because the offline workspace contains only the new files, not a complete Prob4D checkout. Source syntax checks passed. Exact-head repository CI is separate and has not yet been certified.

## Scope and scientific integrity

No new physical recording, dataset download, source residual, protected target, model/checkpoint execution, GPU workflow, or change to historical evidence/claim registries. No BayesianPhysTwin or Causal4D integration benefit is claimed. No stable API/export or provider-v2 semantics are changed. This remains a draft for review, not an auto-merge request.

### DIFF
--- a/CHANGELOG.d/query-information-audit.md
+++ b/CHANGELOG.d/query-information-audit.md
@@ -1,0 +1,7 @@
+Add an experimental complete Gaussian information audit for query-update
+proposals, checking both precision and mean natural parameters against a
+separately retained likelihood. Add absolute query-precision and prior-bounded
+routing metadata, adversarial tests, a registered CPU-only controlled study,
+and independent raw-score verification. No stable API/provider-v2/BPT routing
+or real-provider promotion semantics change. The retained study includes
+fallback accuracy costs and failure under prior shift, not a safety guarantee.
--- a/docs/query-information-audit.md
+++ b/docs/query-information-audit.md
@@ -1,0 +1,191 @@
+# Complete-information audit for query-conditioned updates
+
+Status: **experimental kernel and completed controlled mechanism study**. This
+advances the analytic/controlled parts of issue #339. It does not close the real
+provider qualification or downstream physical-value stages of #333/#49.
+
+## Scientific question
+
+A partially observable reconstruction can constrain a particular query without
+identifying every latent coordinate. However, a useful query covariance does not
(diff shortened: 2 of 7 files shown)


FlorianPfaff:
All exact-head CI checks are green and GitHub now reports this head as mergeable. The connected draft→ready mutation currently fails because its GraphQL query requests an unsupported Repository field. I am closing this draft wrapper and replacing it with an equivalent non-draft PR from the same branch/head so the validated commit can be merged without rewriting history.

PR #347, opened 2026-08-29
Add finite-orbit decision certificates beyond local gauge observability

FlorianPfaff:
## Scientific purpose

Extend the local query-observability work in #338 with a finite-angle, shared-dependence certificate. No existing estimator, facade, frozen evidence, source protocol, or target authorization is changed.

For exact labeled collinear overlap, rotation about the overlap line remains unresolved. The scalar query `q(theta)=cos(theta)` has zero local derivative at the identity despite a full range `[-1,1]`. Under the analytic rank-six factor used in the new integration test, the existing local gate reports direct support 1, variance reduction 10/11, and worst ratio 1/11. An action advantage `0.25+cos(theta)` nevertheless has minimum -0.75.

This is a limitation of a local approximation, not an allegation that the existing API computes its declared local quantities incorrectly.

## Method

- Exact first-harmonic query bounds over a shared axial orbit or closed angle arc, including interior extrema and wraparound.
- Subtract fallback and candidate loss coefficients **before** taking extrema; shared ambiguity can cancel even when independent intervals overlap.
- Require an explicit shared-gauge identity as well as a matching geometric convention. Identical axes do not authorize cancellation of independent latent variables.
- Accept only uniformly positive advantage after a declared omitted-effect envelope, source/model-scope admission, margin, and numerical slack.
- Reject empty anchor support and unadmitted scope; quantify near-line support displacement rather than silently promoting it to exact symmetry.
- Derive the angle arc from one separately admitted Euclidean bounded-error anchor.

The derivation includes the exact SO(2) correspondence stabilizer, a conditional dominance proposition, the squared-distance first-harmonic special case, prior-work positioning, and the downstream ownership boundary. BayesianPhysTwin still owns complete-belief candidate construction and exact fallback.

## Executed controlled mechanism result

Seed 73029; four deliberately constructed families; 512 cases per family, 2,048 total. The orbit and error envelope contain truth by construction.

| Arm | Accepted | Sampled harmful accepts |
|---|---:|---:|
| Local query gate + plug-in comparison | 2,048 | 333 |
| Reject every deficient factor | 0 | 0 |
| Independently bounded query intervals | 512 | 0 |
| Shared finite-orbit certificate | 1,024 | 0 |
| Same certificate + one extra synthetic bounded anchor | 1,659 | 0 |

The equal-information gain is 512 -> 1,024 accepted cases relative to separate intervals. The anchored arm has additional information and is not an equal-budget comparison. The plug-in arm has greater mean deployed advantage than the unanchored certificate in this constructed mixture; this is a harm/acceptance trade-off, not uniform metric dominance. No confidence interval or population-level harm rate is claimed for these engineered controls.

## Validation

Before publication: all **35 standalone tests passed** under Python 3.13.5 / NumPy 2.3.5; source/tests compile; the 2,048-case study was executed. Integration tests against the actual existing query API and lint/type checks are included in the focused hosted workflow, but were not executed in the local partial checkout. CI is authoritative for those checks. The focused workflow also runs the existing query-observability tests and uploads its numerical result.

## Claim boundary

Conditional analytic and constructed-mechanism evidence only. No real provider, protected source, target outcome, physical simulator, or Causal4D outcome was opened. This does not establish real-provider competence, calibrated uncertainty, arbitrary nonlinear-query safety, BayesianPhysTwin benefit, or Causal4D benefit. Floating-point extrema are not outward-rounded interval-arithmetic certificates. Generic rank-six or noisy/near-collinear factors do not prove the required orbit model.

Paper-facing evidence is being retained separately in FlorianPfaff/BayesianPhysTwin-Paper. The decisive future empirical test remains a source-qualified, newly frozen, garment-disjoint PointWorld/Flat'n'Fold physical-query arm under #333, not a retry of a closed negative cohort.

### DIFF
--- a/.github/workflows/finite-orbit-query.yml
+++ b/.github/workflows/finite-orbit-query.yml
@@ -1,0 +1,44 @@
+name: Finite-orbit query mechanism
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - src/prob4d/axial_query_certificate.py
+      - src/prob4d/axial_query_study.py
+      - tests/test_axial_query_certificate*.py
+      - docs/finite-orbit-query-certificates.md
+      - .github/workflows/finite-orbit-query.yml
+
+permissions:
+  contents: read
+
+jobs:
+  mechanism:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+      - name: Install lightweight development dependencies
+        run: python -m pip install -e ".[dev]"
+      - name: Check new source and tests
+        run: |
+          python -m ruff check src/prob4d/axial_query_certificate.py src/prob4d/axial_query_study.py tests/test_axial_query_certificate*.py
+          # Current NumPy stubs use Python 3.12 syntax. Runtime-floor coverage remains in Tests.
+          python -m mypy --python-version 3.12 src/prob4d/axial_query_certificate.py src/prob4d/axial_query_study.py
+          python -m pytest -q tests/test_axial_query_certificate.py tests/test_axial_query_certificate_integration.py tests/test_query_observability.py --junitxml=outputs/finite-orbit-query-tests.xml
+      - name: Execute constructed mechanism only
+        run: |
+          python -m prob4d.axial_query_study --seed 73029 --cases-per-family 512 --output outputs/finite-orbit-query-v1/result.json
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        if: always()
+        with:
+          name: finite-orbit-query-mechanism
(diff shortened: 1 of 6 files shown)


FlorianPfaff:
Paper-side companion is now open: FlorianPfaff/BayesianPhysTwin-Paper#120. It retains the derivation, candidate manuscript wording, complete constructed-study JSON, and source/result hashes without changing existing claims or target-access rules.

Validation update for final head `149b034a72f19a1ae44190fc9a54c2e823245b96`:

- Full Tests workflow `33279982174`: **success**.
- Finite-orbit query mechanism `33279982150`: **success** (Ruff, new-module mypy, unit/integration/regression tests, and the 2,048-case study process).
- Fail-closed quality `33279982170`: **success**.
- Security scanning `33279982163`: **success**.
- Provider batch preflight `33279982217`: **success**.

The local retained result SHA-256 is `321ff79cc7446b345da884da8d8e893114fc822fda54b5c679bc52fc4f7e30c4`. All six local source Git-blob/SHA-256 identities and this result digest were verified against the paper manifest. The local standalone test XML records 35 passed, zero failed. Hosted tests additionally execute the five new existing-API integration cases. No bitwise equivalence between local and hosted numerical JSON is claimed.

Both PRs remain unmerged. These checks establish implementation and constructed-mechanism evidence, not real-provider competence, BPT physical benefit, Causal4D validity, or deployment safety.

PR #349, opened 2026-08-30
Add shared-gauge curvature moments for tangent-null physical queries

FlorianPfaff:
## Scientific objective

Add an experimental, additive uncertainty representation for a specific nonlinear blind spot in shared `Sim(3)` gauge propagation.

For the controlled query

`q = ell * sin(alpha) * sin(beta)`

with independent zero-mean Gaussian rotations, both first-order propagation and the existing axis spherical-radial reference can report zero query variance even though the true variance is positive. This PR adds a quadratic Gaussian-moment representation that preserves the resulting curvature covariance *jointly* across outputs rather than treating it as independent pointwise noise.

This work complements the query-aware observability program in #339: observability decides whether information supports a declared physical query; this PR addresses a distinct case where the nonlinear query has leading uncertainty outside the first-order tangent response.

## Implementation

- `SharedGaugeMoments`: mean plus shared linear/curvature covariance factors, covariance actions, query projection, and input cross covariance.
- `quadratic_gaussian_moments`: exact Gaussian moments for a local quadratic surrogate.
- `finite_difference_gauge_moments`: centered stencil including all mixed second derivatives; `1 + 2 r^2` evaluations; hard rank cap rather than silent truncation.
- `sim3_chain_gauge_moments`: applies the method to the existing Prob4D `Sim3` chain and optionally projects directly to a fixed physical query.
- focused analytic/upstream-Sim3 regression tests and a deterministic controlled study.

The stable `prob4d.api.v2` facade, CLI registry, artifact schemas, provider-v2 exporter, admission rules, and existing gauge-linearization closure diagnostic are unchanged.

## Controlled finding

At `ell=0.5 m`, `sigma=0.1 rad`:

- exact query SD: `4.950331673 mm`;
- first-order SD: `0`;
- axis spherical-radial SD: `0`;
- quadratic shared-curvature SD: approximately `5.0 mm`.

In the fixed scalar physical-update analogue with 64 observations, 2 mm prior SD, and 1 mm independent point noise, the shared-curvature arm has expected RMSE `1.854503 mm` versus `2.000000 mm` for exact physical fallback, a `7.27%` reduction. The much larger improvement versus the deliberately unguarded first-order arm is not a claim about deployed BayesianPhysTwin.

Strong comparators are retained: fifth-degree cubature and 3-node tensor Gauss-Hermite are more accurate on the rank-two sine example; at `sigma=0.35 rad` the local quadratic approximation overestimates the example variance by `27.12%`.

## Validation already performed on the exact base

This branch starts at `224d13dc9a93731ac5297b479eb1e121b3dbe659`, the exact Prob4D revision used by the controlled study. The prepared patch passed 50 focused tests locally and a repeated study run produced byte-identical outputs/manifests. GitHub CI should independently validate the pushed branch.

## Evidence and claim boundary

Controlled synthetic mechanism evidence only. This PR does **not** establish:

- real-provider competence or transfer;
- calibrated real uncertainty;
- BayesianPhysTwin guard benefit;
- Causal4D intervention benefit;
- a new state-of-the-art result;
- globally conservative nonlinear uncertainty.

Quadratic Gaussian moments and higher-degree cubature are established prior art. The contribution candidate is the shared-query uncertainty semantics, the identified tangent-null/axis-rule blind spot, and a testable downstream consequence.

Paper-facing evidence is being submitted separately to `FlorianPfaff/BayesianPhysTwin-Paper` so code and scientific interpretation remain independently reviewable.

Refs #339 #49

### DIFF
--- a/docs/gauge-curvature.md
+++ b/docs/gauge-curvature.md
@@ -1,0 +1,221 @@
+# Shared-gauge curvature for physical 4D queries
+
+**Status:** experimental, additive numerical method; not a provider-v2 exporter,
+calibration certificate, or change to any existing admission rule.
+
+## Motivation and the exact boundary of the finding
+
+Prob4D's source-only closure diagnostic compares analytic first-order propagation
+with a spherical-radial rule evaluating `+/- sqrt(r) L[:, j]`. The relevant
+upstream implementation is `_sigma_points` in
+`src/prob4d/_gauge_linearization_numerics.py`, inspected at commit
+`224d13dc9a93731ac5297b479eb1e121b3dbe659` (blob
+`9dd800dbdd9b2f262eec0a50ddc3aaa2d871873c`). See the existing
+[gauge linearization closure guide](gauge-linearization-closure.md).
+
+Those two approximations can agree on zero query uncertainty even when the
+nonlinear query has positive variance. This is a query-level blind spot, **not**
+a demonstrated false pass of the entire existing closure/admission workflow.
+Other point-level, mean-shift, support, or physical guards may still reject.
+The original diagnostic and its frozen artifacts are intentionally unchanged.
+
+Take a point `(0, 0, ell)` and two independent zero-mean Gaussian angles with
+variance `sigma^2`. Compose `Rz(alpha) Ry(beta)` and query the y coordinate:
+
+\[
+q = \ell\sin\alpha\sin\beta,\qquad
+\mathbb E q=0,\qquad
+\operatorname{Var}(q)=\ell^2
+\left(\frac{1-e^{-2\sigma^2}}{2}\right)^2.
+\]
+
+The first derivative at zero vanishes. Every axis sigma point also makes one
+angle zero. Both approximations therefore report zero query variance. With
+`ell = 0.5 m` and `sigma = 0.1 rad`, the exact standard deviation is
+`4.950331673 mm`. A multiplicative covariance inflation cannot recover a
+missing direction whose original covariance is exactly zero.
+
+## Classical quadratic moments, represented jointly
+
+For a *complete joint* root `L`, write `g = mu + L z`, `z ~ N(0, I_r)`. Let a
+quadratic surrogate for each output coordinate be
+
+\[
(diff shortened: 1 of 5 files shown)


FlorianPfaff:
## DOT rope source-only real-geometry result

The installed DOT rope release was sufficient to run a first real-geometry stress test. A separate retained execution completed on branch `research/dot-rope-real-pilot-v1`, Actions run `33302466278`, with artifact ID `bf6c1439a3f61416c47c01dc441f0ae9215b5ca2eb8c8ba07aff3b17ed4df885`.

### Information boundary

- opened only publisher 3-D coordinate members for `R01`, `R02`, and `R03`;
- no normal/UV images, 2-D correspondences, learned provider, BayesianPhysTwin, or Causal4D execution;
- `R04-R70` remain reserved and must not be opened without a new frozen authorization;
- the angular prior/noise model is controlled rather than estimated from a real provider.

### Result

Across 179 sampled real rope geometries, 114 met the frozen near-linear/off-axis primary geometry criterion. At the primary observation-noise level, local first-order propagation passed the scalar uncertainty screen while the finite posterior screen failed in **114/114** primary cases.

For those 114 cases, mean normalized expected Gaussian NLL was:

- posterior circular moments: `0.609500`;
- 32-node Gauss-Hermite: `0.620969`;
- five-node Gauss-Hermite: `0.620302`;
- local second order: `0.659523`;
- local spherical-radial-2: `14950.800980`;
- local first order: `25327.070885`.

The local second-order approximation was directionally useful but substantially less accurate than finite-orbit/circular propagation: mean absolute normalized variance error `0.626391` versus approximately numerical zero for posterior circular moments. Thus DOT supports the *failure mechanism* on real rope geometries, but at the deliberately broad 1-rad axial prior it favors an exact finite-orbit/circular representation over treating the quadratic approximation as the final method.

### Claim boundary

This is real-geometry controlled-gauge evidence, not real-provider competence, empirical uncertainty calibration, guarded BayesianPhysTwin benefit, Causal4D benefit, or held-out confirmation. It strengthens the paper direction while narrowing the technically strongest method claim: local shared curvature is a small-uncertainty closure; finite-orbit moments are the stronger continuation for large rotational uncertainty.

PR #350, opened 2026-08-30
Add nonlinear circular-gauge query propagation

FlorianPfaff:
## Purpose

Add an experimental method kernel for **nonlinear physical-query uncertainty under a declared unresolved circular gauge**. The contribution is deliberately narrower than a new provider adapter: it asks whether retaining a rank-deficient 4-D gauge is sufficient when a downstream physical query is nonlinear in the unobserved direction.

This is additive and does not modify provider-v2, the stable API, existing estimators, admission decisions, or historical evidence.

## Method

For a fixed observable quotient and an explicitly justified conditional circular prior, the new module represents a complete stacked query as

```text
q(phi) = c + a cos(phi) + b sin(phi)
```

with **one shared phase** across points and time. It provides:

- exact conditional first and second moments from circular harmonics;
- a two-column covariance factor preserving all cross-point/cross-time dependence in linear memory;
- exact union-of-arcs geometry for affine path constraints;
- wrapped-normal mixture event probabilities with an explicit omitted-normal-tail bound; and
- a model-conditional risk predicate whose caller still owns exact fallback.

A local rank-deficient Jacobian is explicitly **not** treated as proof of a global circular symmetry. The caller must justify the likelihood invariance and retain the conditional phase prior rather than inventing a uniform/zero-twist replacement.

## Controlled analytic result

For a 100 mm off-axis probe with unresolved phase `phi ~ WN(0, 1 rad^2)` and the illustrative event `x < 50 mm`:

| Representation | radial mean | radial std. | P(x < 50 mm) |
| --- | ---: | ---: | ---: |
| first-order Gaussian | 100.000 mm | 0.000 mm | 0.000% |
| rank-one `+/- sigma` spherical-radial | 54.030 mm | 0.000 mm | 0.000% |
| 17-node Gauss-Hermite, direct indicator | 60.653 mm | 44.698 mm | 24.705% |
| exact-moment Gaussian | 60.653 mm | 44.698 mm | 40.581% |
| exact circular shared phase | **60.653 mm** | **44.698 mm** | **29.501%** |

An independent 200,000-draw Monte Carlo control gives 29.601% violation, with a 95% Wilson interval containing the analytic value.

This is a counterexample to the **specified local and rank-one two-point rules**, not a claim that all sigma-point methods fail. Higher-order quadrature recovers the smooth moments; matching moments alone still does not certify a discontinuous event probability.

## Shared-phase dependence control

Five constraints with identical 10% marginal phase risk have 10% joint risk when they are the same event. Five disjoint phase arcs with the same marginals have 50% joint risk. Treating the five constraints as independent gives `1 - 0.9^5 = 40.951%` in both cases, so the independence approximation can be either conservative or anticonservative.

## Files

- `src/prob4d/circular_gauge_query.py` — experimental runtime kernel;
- `tests/test_circular_gauge_query.py` — analytic, adversarial, geometry, probability, and dependence tests;
- `scripts/science/circular_gauge_query_study.py` — deterministic analytic/synthetic study;
- `scripts/science/verify_circular_gauge_query_evidence.py` — independent standard-library verifier that imports no Prob4D code;
- `protocols/circular-gauge-query-control-v1.json` — frozen mechanism-study specification;
- `docs/circular-gauge-query.md` — derivation, assumptions, claim limits, and prospective empirical path.

## Validation

The exact branch bytes were checked against the locally validated files. On those bytes:

```text
PYTHONPATH=src python -m pytest -q tests/test_circular_gauge_query.py
36 passed in 1.17s
```

The previously generated evidence bundle was additionally checked by the independent standard-library verifier for analytic values and source/evidence hashes. Full repository CI, installed-wheel integration, real-provider execution, and provider target scoring are intentionally left to the PR/repository workflows.

## Information and claim boundary

This study opens no real source or target data and runs no real provider, BayesianPhysTwin, or Causal4D inference. It establishes a bounded mechanism/theory result only. It does **not** establish:

- real-provider competence or unseen-object benefit;
- calibrated real-world uncertainty;
- a complete nonlinear Sim(3) posterior;
- that every local gauge nullspace is a global circular symmetry;
- BayesianPhysTwin or Causal4D empirical improvement;
- robot-control safety; or
- state of the art.

The PointWorld/Flat'n'Fold path in #333 remains independently gated. This kernel must not authorize target access or rescue a failed provider gate. It becomes paper-material only if source-side evidence supports a material residual circular ambiguity for the registered query; otherwise it remains a bounded diagnostic/theory extension.

Tracks the method-level objective of #49 without changing its provider-competence decision order.

### DIFF
--- a/docs/circular-gauge-query.md
+++ b/docs/circular-gauge-query.md
@@ -1,0 +1,261 @@
+# Nonlinear physical queries under an unresolved circular gauge
+
+**Status:** experimental, locally implemented and tested on 2026-08-30. Additive
+patch reviewed against Prob4D `224d13dc9a93731ac5297b479eb1e121b3dbe659`.
+No stable API, existing estimator, admission decision, or historical evidence is
+changed. This is not a real-provider or deployment result.
+
+## Why this is a scientific extension rather than another adapter
+
+The existing `observable_gauge` path retains a rank-deficient Sim(3) information
+factor. `query_observability` projects that factor with a local query Jacobian.
+The separate gauge-linearization-closure diagnostic compares first-order
+propagation with a spherical-radial reference using `2r` symmetric points for a
+rank-`r` Gaussian coordinate covariance. These are useful, explicitly local
+instruments, but neither a local derivative nor that specific cubature rule
+establishes correctness for a nonlinear physical-query distribution.
+
+A circular ambiguity gives a rigorous example. Rotation about a line can be
+unobserved by collinear geometric correspondences. An off-axis point then lies
+on a circle, not on a Gaussian tangent line. In particular, a radial coordinate
+can have zero local derivative and strictly positive variance. The symmetric
+rank-one two-point cubature rule also misses that variance exactly.
+
+The extension propagates an **explicit conditional circular prior** through
+such queries, preserving a single shared phase across points and time. It
+provides exact conditional moments in a two-column covariance factor and
+joint linear-constraint event probabilities through unions of circular arcs.
+
+## Premises: local nullspace is not a global symmetry certificate
+
+Let `psi` denote the observable quotient and `phi` a residual rotation. The
+required likelihood property is the GLOBAL conditional invariance
+
+    L(y | psi, phi) = L(y | psi) for every supported phi.
+
+Under that premise,
+
+    p(phi | psi, y) = p(phi | psi).
+
+The conditional prior must be retained, not replaced by a uniform density, a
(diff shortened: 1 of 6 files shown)


FlorianPfaff:
Closing as the lower-priority nonlinear-gauge alternative after #349 landed. This draft requires a globally justified residual circular phase model, has no real-provider evidence, and still fails the fail-closed Ruff gate (48 style diagnostics). The exact circular-event machinery can be revived as a focused follow-up if a real source-side residual ambiguity actually supports the circular-symmetry assumption.

PR #355, opened 2026-08-30
Add target-closed Deform360 source-bundle audit

FlorianPfaff:
## Purpose

Begin the fresh Deform360 validation path with a metadata-only inventory of the existing processed source bundle on the self-hosted runner labelled `gpuserver4090`.

The run targets exactly:

`/home/florianpfaff/deform360-fresh-source-processed-v1-1a3f9b1`

## Safety and evidence boundary

- metadata only: directory-entry names, types, mode bits, and regular-file sizes;
- no dataset file-content reads;
- no provider execution, predictions, residuals, covariance fitting, or target access;
- no dataset mutation;
- symlinks are recorded but never followed;
- target-like path components are skipped before stat or descent;
- evidence is written only under `RUNNER_TEMP` and uploaded from a read-only self-hosted job;
- a separate hosted job owns the issue-comment write token.

The metadata digest is deliberately not a byte-complete dataset manifest and establishes no provider competence or scientific performance.

## Control plane

The reviewed workflow is triggered after merge only when the single file `protocols/execution_requests/deform360_source_bundle_audit_v1.json` changes in a non-forced push to `main`. The request must bind the exact Git blob of the source protocol and its own canonical SHA-256 identity.

The execution job uses `runs-on: [self-hosted, gpuserver4090]` and the existing `trusted-self-hosted-validation` environment.

## Validation

Locally validated against the current repository with:

- Ruff check and formatting;
- focused source-audit tests;
- trusted self-hosted workflow policy tests;
- GitHub Action pin tests;
- Python compileall; and
- YAML parsing and whitespace checks.

Related: #49 and #348.

### DIFF
--- a/.github/workflows/deform360-source-bundle-audit.yml
+++ b/.github/workflows/deform360-source-bundle-audit.yml
@@ -1,0 +1,389 @@
+name: Deform360 source-bundle audit
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/deform360-source-bundle-audit.yml"
+      - "scripts/science/audit_deform360_source_bundle.py"
+      - "protocols/deform360-source-bundle-audit-v1.json"
+      - "tests/test_deform360_source_bundle_audit.py"
+      - "tests/test_deform360_source_bundle_workflow_policy.py"
+      - "tests/test_trusted_self_hosted_validation_policy.py"
+      - "docs/deform360-source-bundle-audit.md"
+  push:
+    branches: [main]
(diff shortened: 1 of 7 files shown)


FlorianPfaff:
Closing as an exact duplicate/superseded PR. PR #353 already merged the same Deform360 source-bundle audit from the same head SHA `6c8a32407549d2b468e81fe6aea9fd75f4521a5c`. Keeping #355 open only creates a stale conflict against `main` without adding any code or evidence.

PR #342, opened 2026-08-29
Scientific kernel: posterior-preserving compression of shared observation noise

FlorianPfaff:
## Research question and contribution
Can shared 4D observation uncertainty be compressed while preserving the **conditioned physical-query posterior**, rather than only a marginal covariance projection?

For a frozen local Gaussian model with full innovation covariance `S = A + U U.T`, `A > 0`, and query/innovation cross covariance `C`, retaining `range(U.T solve(S, C.T))` preserves the query gain for every innovation and its posterior covariance. The documentation proves necessity, sufficiency, minimal rank `rank(U.T solve(S,C.T)) <= query_dimension`, and an explicit covariance/mean-error identity for truncation.

This extends the existing marginal query compressor and structured query conditioner; it does not claim either existing module violates its documented contract. If the marginal projection is already the full Bayesian gain, its exact preservation gives the same subspace condition. Goal-oriented inference is established prior work (Spantini et al., 2017, DOI 10.1137/16M1082123); no first-ever or SOTA claim is made.

## Retained failure / first readiness boundary
An analytic counterexample retains 99.9999000001% of shared covariance trace and an exact declared marginal projection, yet understates conditioned scalar variance by 5000.7475x. The new rank-one factor preserves it. This is an algebraic conditioning failure, **not a localized real-provider failure**. No provider admission/readiness gate changes and no new point-covariance family, target guard, adapter, or calibration is introduced.

## Implementation
- Experimental NumPy-only `posterior_preserving_compression` module; no stable API or claim-bearing export changes.
- Batched existing-operator solve protocol and posterior-whitened numerical subspace selection.
- Downdate audits of gain, posterior covariance and expected normalized mean error.
- Exact original-factor fallback if the rank cap cannot preserve the posterior; no ridge or forced truncation.
- Proof, counterexample, independent tests, existing-operator integration test, and fixed controlled Sim(3)-linearized protocol.

## Validation actually performed
Local Python 3.13.5 / NumPy 2.3.5:
- `PYTHONPATH=src OPENBLAS_NUM_THREADS=1 python -m pytest -q tests/test_posterior_preserving_compression.py`: **52 passed**.
- Syntax compilation passed for the module, study script and integration test.
- The full repository suite, Ruff, and existing-operator integration test were **not run locally**: this workspace contains the new files, not a complete checkout. The integration test is included for repository CI; do not interpret its existence as a pass.

[...]

PR #375, opened 2026-08-30
Evaluate posterior-preserving compression on incomplete PokeFlex real geometry

FlorianPfaff:
## Purpose

Run the posterior-preserving shared-noise compression from merged PR #342 on the available real PokeFlex geometry at `/mnt/lexar4tb/pokeflex`, without treating the reported 170 ZIPs / ~548 GiB as a complete release.

This is a separately versioned **real-geometry local-Gaussian mechanism diagnostic**. It does not continue, repair, retune, or open the prior Causal4D PokeFlex target experiment, and it does not promote a real provider.

## Data handling and selection

- Inventory every observed ZIP central directory, member suffix distribution, size, and CRC metadata.
- Record corrupt, encrypted, oversized, unsupported, and identity-ambiguous geometry explicitly.
- Deep-read at most 24 archives selected by ascending SHA-256 of relative archive path plus candidate key; no error or covariance result participates in selection.
- Read only non-pickled NumPy trajectories or stable-vertex OBJ/PLY/XYZ/PTS sequences. Geometry member reads receive normal ZIP CRC verification.
- Never persistently extract archives or upload raw data.

## Real-geometry experiment

For each safely admitted trajectory, construct a local Gaussian stress model from real prefix material-point motion and a cumulative Sim(3) gauge chain. Register a 3-D endpoint-patch displacement query and compare:

1. full shared covariance;
2. posterior-preserving rank <= 3;
3. equal-rank observation-covariance PCA;
4. conditional-only uncertainty; and
5. the cached full query message.

Primary endpoints are full/compressed gain, posterior covariance and realized posterior-mean parity, retained rank, and shared-factor payload ratio. NLL/NEES and endpoint errors are diagnostic only; the fitted local model is not claimed calibrated.

## Corrected execution binding

The mirror is physically hosted on `gpuserver6000`, not `gpuserver4090`. Commit `33aeaac22c81e88afe6eb9cc1aedbf55919dfe41` binds the request and workflow to `[self-hosted, Linux, X64, gpuserver6000]`, verifies the canonical dataset root before checkout, and triggers the run through the request-file change. The earlier wrong-runner preflight did not open data.

The data job is restricted to actor `FlorianPfaff`, this repository and exact branch/SHA, read-only permissions, no persisted checkout credentials, no repository secrets, and the `trusted-self-hosted-validation` environment. Compact evidence is uploaded even when the dataset layout is not evaluable; the final job then fails rather than relabeling such a state as a scientific negative.

## Claim boundary

A positive run establishes numerical posterior preservation and query-sufficient rank on a deterministic bounded subset of the observed real geometry under the stated local Gaussian model. It does **not** establish PokeFlex completeness, real-provider competence, real covariance calibration, BayesianPhysTwin physical benefit, Causal4D benefit, independent-object generalization, deployment safety, or state of the art.


FlorianPfaff:
Closing during PR consolidation: this branch is an incomplete PokeFlex geometry diagnostic built on a non-main science base and has no completed outcome-bearing result. The available PokeFlex mirror is incomplete, so keeping this open adds maintenance/conflict surface without paper-level evidence.

PR #384, opened 2026-08-30
Fix sealed CUT3R checkpoint loading under PyTorch 2.11

FlorianPfaff:
## Cause

The retained runtime smoke reached the pinned CUT3R checkpoint but failed because PyTorch 2.11 defaults `torch.load` to `weights_only=True`. The frozen CUT3R loader omits that argument, and the pinned checkpoint contains an OmegaConf `DictConfig`.

## Repair

- Activate PyTorch's documented `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` compatibility mode only after the exact native-RoPE runtime artifact and exact checkpoint SHA-256 both match the sealed identities.
- Fail closed on checkpoint mismatch or a conflicting `TORCH_FORCE_WEIGHTS_ONLY_LOAD` policy.
- Add focused regression tests for activation, nonmatching runtimes, hash mismatch, and conflicting policy.

The DOT source/held-out split, provider, checkpoint, native CUDA runtime, and scientific thresholds are unchanged.

### DIFF
--- a/src/prob4d/cut3r_runtime_contract.py
+++ b/src/prob4d/cut3r_runtime_contract.py
@@ -6,6 +6,7 @@ import hashlib
 import importlib
 import importlib.machinery
 import json
+import os
 import sys
 from collections.abc import Mapping
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
Closing as superseded by merged PR #382, which fixes the same PyTorch 2.11 checkpoint-loader failure with a stricter exact-source/preimage rewrite and a content-addressed compatibility receipt. The merged-main source evaluation is running as workflow 33329050649.

PR #385, opened 2026-08-30
Fix DOT CUT3R runtime-smoke temporary directory collision

FlorianPfaff:
## Cause

Repaired workflow run `33329050649` successfully loaded the exact pinned CUT3R checkpoint, instantiated `ARCroco3DStereo`, and matched all model keys. It then stopped before synthetic inference because `tempfile.TemporaryDirectory()` had already created its root, while `_make_synthetic_frames()` was asked to create that same directory with `exist_ok=False`.

## Repair

- pass a fresh `frames/` child of the temporary root to `_make_synthetic_frames()`;
- retain the helper's fail-closed requirement that its destination must not already exist;
- add regressions for the exact runtime-smoke call site and for creating three synthetic frames under a fresh child directory.

This does not change CUT3R, the checkpoint, CUDA/native-RoPE identities, DOT sequence selection, provider equations, uncertainty closures, scoring, or the marker-separation boundary. `R04-R70` remain unopened.


FlorianPfaff:
Closing as superseded by merged PR #386. PR #386 implements the same child-workspace correction in the isolated, request-bound runtime preparation and additionally binds the exact provider-script blob/preimage with a compatibility receipt. Merging this direct source edit would change that registered blob and invalidate the stricter sealed-runtime guard. The request-only main commit 7eb4867e36742d819c514fad21436d4f475b4bed is now executing the repaired evaluation.

PR #388, opened 2026-08-30
Audit DOT marker support against the sealed CUT3R provider

FlorianPfaff:
## Purpose

Continue the real DOT source experiment after workflow run `33329701704` established a valid marker-free CUT3R provider bundle but the separate marker evaluation stopped with `ValueError: fewer than six valid marker samples remain`.

This PR does not change the provider, coordinate convention, support threshold, uncertainty closures, scoring code, source roster, or scientific claim. It adds a separately authorized source-only audit to determine whether the failure is caused by the six-markers-per-frame implementation gate, a coordinate-column/unit mismatch, or genuinely insufficient marker support.

## Frozen inputs

- official DOT V29 archive: `R01-10.zip` under `/mnt/seagate10tb/florianpfaff/datasets/dot`;
- source sequences only: `R01`, `R02`, `R03`;
- reserved and unopened: `R04-R70`;
- immutable provider workflow run: `33329701704`;
- provider artifact: `dot-rope-cut3r-sealed-provider-33329701704-1`;
- provider bundle ID: `952421d140731b2a6eb99df3cbd348653e04863fa457aaa490be31fe0b4c06a7`.

## Audit

The self-hosted job runs on `[self-hosted, Linux, X64, gpuserver4090]`, downloads the exact previously sealed provider artifact, verifies its content-addressed manifest and member digests, then opens only the R01-R03 2-D/3-D marker files. It tests explicit pixel, one-based, unit-normalized, and percent-normalized coordinate hypotheses over the available numeric columns and records per-frame and pooled support.

It computes **no reconstruction or uncertainty performance metric** and cannot select a method from outcomes. Terminal decisions are limited to:

1. current convention is pooled-support feasible but the per-frame gate is too strict;
2. current convention is already sufficient and another implementation defect remains;
3. exactly one alternative convention is required;
4. the coordinate contract is ambiguous; or
5. source marker support is genuinely insufficient.

## Trigger and custody

The implementation is reviewed on hosted CI. After merge, a separate commit must change only `protocols/execution_requests/dot_rope_marker_support_audit_v1.json`; the authorization job rejects multi-file, forced, deleted, malformed, or identity-drifted triggers. The self-hosted job has read-only repository and Actions permissions and uploads only aggregate diagnostics plus source-member hashes—never images, raw coordinates, or reserved payloads.

## Local validation

Focused request, parser, coordinate-transform, pooled-support, and protected-workflow tests pass locally. Repository CI remains authoritative.

Refs #49 #386


FlorianPfaff:
Closing during PR consolidation. This PR is a source-only marker-support diagnostic/audit and explicitly computes no reconstruction or uncertainty performance metric. It is useful troubleshooting infrastructure, but it does not add a retained paper-level result and would keep the open queue alive after the completed Deform360/DLO evidence has been consolidated.

PR #402, opened 2026-08-31
Measure query-portfolio factor/cache break-even on real cloth trajectories

FlorianPfaff:
## Scientific question

When is a posterior-preserving shared-noise factor actually preferable to caching the Gaussian query gain and posterior covariance directly?

For one immutable query, the cached message is the simpler and generally faster baseline. This PR makes that objection an explicit real-data experiment rather than hiding it.

## Experiment

Use the complete public Tracking Cloth Deformation dataset in five recording-disjoint folds per cloth size. Fit one source-side joint Gaussian model for all marker future displacements and the causal marker-motion observation, then register deterministic portfolios of 1, 2, 4, 8, 12, and (for A2) 20 marker queries.

For every portfolio, report:

- exact full/compressed gain, posterior covariance, and realized-mean parity;
- retained shared rank;
- factor payload, factor-plus-projection payload, and direct joint-query `(K, P)` cache payload;
- construction time and per-window update time using a block-diagonal-plus-low-rank Woodbury solver; and
- a positive-definite generic comparator formed by a low-rank-plus-isotropic spectral approximation of the **complete joint covariance**.

The payload comparison credits the factor only when `Q`, `C`, conditional covariance, and means are already resident at the consumer. The protocol explicitly declares direct caching preferable when those blocks are not resident or when minimal online latency for one immutable query dominates.

## Validation state

Hosted run `33364480389`, job `99402284665`, completed on the official Zenodo release. Artifact `9747804017` has ZIP SHA-256 `310775ea119a86e9728f2b78e82cb46e7b86fb9c657cfbb5a1237278586a8efc`; all nine members were independently rehashed. It covers 80 accepted recordings, five recording-disjoint folds, 55 fold/portfolio evaluations, and 7,648 unique held-out windows.

Exact compression reproduced the full posterior with maximum relative gain error `4.0804182725523995e-12`, maximum relative covariance error `5.383026917575375e-13`, and maximum realized mean difference `5.797456725464489e-13 m`. The shared factor alone used 50.0%-95.24% of the joint-cache payload, but factor plus query projection was never smaller for the registered finite portfolios. Cached updates were approximately 13.5-17.0 times faster than reconstructing from compressed factors.

The compact receipt is archived in `evidence/tracking-cloth-query-portfolio-v1/` with summary SHA-256 `3080195fb726478322c70a203db5871221017fc8283c5ca9bd38ff006b30081b`. The separate local-mirror job `99402284714` remains untouched and is not claim evidence. The mainline policy conflict was resolved by preserving the DOT held-out confirmation and both Tracking-Cloth workflow registrations; the focused policy/query/compression suite passes 73/73.

## Boundaries

This is a recording-disjoint real-motion-capture representation/cost study. It does not establish learned 4D-provider competence, BayesianPhysTwin physical benefit, Causal4D benefit, deployment calibration, or state of the art. Raw trajectories are never copied into the uploaded evidence.


FlorianPfaff:
## Official real-data result (checksum-verified Zenodo execution)

The file-triggered hosted run `33364480389` completed successfully at exact source revision `3dfe9726b534f2dd556a706afeaaf4e69027ccf4`. It downloaded the official 27.4 MB `tracking_dataset.zip`, verified publisher MD5 `b4868b702f8a42b2ea1069d0f1a3b8f6`, found 120 CSV recordings, admitted the same 80 cloth-only recordings, removed the downloaded archive/trajectories before artifact upload, and passed the exact-posterior enforcement step.

Artifact: `9747804017`, digest `sha256:310775ea119a86e9728f2b78e82cb46e7b86fb9c657cfbb5a1237278586a8efc`.

Evidence identities:
- protocol SHA-256: `798903446c5de29c8bee54e83a0d9c63f2edf375c8419f626d3867d15aed6491`
- result SHA-256: `4361d7dee3adf36c51cf4d0687dcfdd5d70cfa2323adc00ae839bdc10fc9e680`
- dataset manifest SHA-256: `e3f4bda9da258af6717ef9e5cbbd8c0e582ad2f80efe8419ca4c853d377a05c0`
- Python 3.12.14 / NumPy 2.3.5
- raw data copied to evidence: false

### Exactness and rank
Across 55 fold/portfolio evaluations (five recording-disjoint folds per size and portfolios of 1, 2, 4, 8, 12, and where available 20 marker queries):
- maximum relative gain error: `4.0804e-12`
- maximum relative posterior-covariance error: `5.3830e-13`
- maximum realized posterior-mean difference: `5.7975e-13 m` (`5.7975e-10 mm`)
- every exact posterior valid; every PSD spectral baseline posterior valid
- retained rank equalled the joint query dimension: `3, 6, 12, 24, 36`, and `60`
- the complete-marker portfolios therefore correctly returned exact full-rank fallback in all five folds for A2 (`60/60`) and A3 (`36/36`)

### Factor versus direct query cache
Under the registered scenario where `Q`, `C`, conditional covariance, and means are already resident, the reduced-factor payload was smaller than the joint `(K, P)` cache in every fold:
- A2 factor/cache ratio: `0.952, 0.909, 0.833, 0.714, 0.625, 0.500`
- A3 factor/cache ratio: `0.923, 0.857, 0.750, 0.600, 0.500`

This is a modest 4.8--7.7% factor advantage for one marker and a 50% advantage at the complete-marker portfolio. Retaining the query-specific projection metadata removes most or all of that advantage; it is not required by an online consumer once the reduced factor has been materialized.

The direct cache decisively wins online latency:
- cached update: `0.036--0.207 us/window`
- compressed-factor update: `0.519--2.897 us/window`
- full-factor update: `1.415--2.913 us/window`

Thus compression accelerates the structured factor solve by up to about `3.48x` for small portfolios, but cached matrix multiplication remains roughly `13--16x` faster than the compressed-factor path. Compression construction (`0.586--1.413 ms`) is also slower than direct cache construction (`0.096--0.280 ms`).

### Valid generic comparator
The low-rank-plus-isotropic spectral approximation of the complete joint covariance stayed positive definite in every fold, but increased per-coordinate RMSE relative to the exact full model. Examples:
- A2, one marker: `6.304 -> 9.077 mm`
- A2, 20 markers: `7.364 -> 8.981 mm`
- A3, one marker: `6.109 -> 6.898 mm`
- A3, 12 markers: `5.441 -> 6.777 mm`

### Scientific interpretation
This resolves rather than evades the cache objection. For one immutable query and latency-sensitive online inference, cache `(K,P)`. The factor representation has a bounded storage advantage only when the surrounding Gaussian model is already resident and a factor-consuming interface is required. The real portfolios also show that the rank bound can be tight: independent 3-D marker queries add three required directions each, and no reduction remains for the complete marker state.

The `gpuserver4090` run remains queued as an independent local-mirror reproduction; no claim depends on it.

PR #404, opened 2026-08-31
Add held-out real-DLO evidence for query-aware observability

FlorianPfaff:
This is an idempotent lookup attempt for the existing draft PR; no duplicate should be created.


FlorianPfaff:
Closing as redundant/accidental PR scaffolding. The PR body explicitly states that this was an idempotent lookup attempt and that no duplicate should be created. The underlying scientific result is retained in the proper DEFORM query-observability PR lineage, so resolving this draft's merge conflict would add no value.

PR #434, opened 2026-08-31
Run frozen DOT R11–R30 confirmation on gpuserver6000

FlorianPfaff:
## Purpose

Move only the learned-provider execution lane of the already frozen DOT R11–R30 query-selective experiment from the unavailable `gpuserver4090` runner to the prepared `gpuserver6000` runner.

## Scientific invariants retained

- unchanged protocol: `protocols/dot-rope-query-selective-heldout-v1.json`;
- unchanged R04–R10 strong-positive prerequisite;
- unchanged target cohort R11–R30 and untouched reserve R31–R70;
- unchanged CUT3R revision and checkpoint;
- unchanged methods, observability thresholds, query definitions, exact fallback, sequence-level bootstrap, comparators, and terminal classifications;
- unchanged prediction-first information order: normal-view provider seal, then 2-D-marker factor/prediction seal, then one-time 3-D scoring.

## Operational change

The provider job now binds `[self-hosted, gpuserver6000]` / `workstation2` and consumes the separately attested persistent CUT3R runtime plus publisher-verified compressed archives. Hosted jobs still independently download and verify the official archives before factor sealing and outcome scoring.

## Fail-closed behavior

The workflow rejects any request not bound to an exact `heldout-strong-positive` R04–R10 artifact. It rejects changed runtime/provider identities, opened prewarm archives, marker access during provider inference, malformed seals, or unsupported terminal results. No target-side tuning or downstream BayesianPhysTwin/Causal4D execution is authorized.

### DIFF
--- a/.github/workflows/dot-rope-query-selective-heldout-gpuserver6000-v2.yml
+++ b/.github/workflows/dot-rope-query-selective-heldout-gpuserver6000-v2.yml
@@ -1,0 +1,840 @@
+name: DOT rope query-selective held-out CUT3R gpuserver6000 v2
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/dot-rope-query-selective-heldout-gpuserver6000-v2.yml"
+      - "CHANGELOG.d/dot-query-selective-gpuserver6000-v2.md"
+      - "tests/test_dot_rope_query_selective_gpuserver6000_v2_workflow.py"
+  push:
+    branches: [main]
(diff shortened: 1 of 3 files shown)


FlorianPfaff:
Closing as superseded. PR #435 explicitly replaces this first operational gpuserver6000 draft with the corrected/testable R11–R30 control plane while preserving the frozen scientific protocol. Keeping this older unmergeable control-plane draft open adds no distinct scientific or operational value.

PR #465, opened 2026-09-01
Fix DOT recovery control request packaging

FlorianPfaff:
The repaired R04–R10 recovery reached and independently classified the frozen scientific result as `heldout-support-negative`, but evidence packaging failed because the recovery job had checked out only the historical evaluator revision under `frozen/` while reading and copying the current recovery request from the nonexistent workspace-root path.

This change:
- checks out the exact merged recovery-control revision separately under `control/`;
- reads and revalidates the content-addressed recovery request from that explicit checkout;
- copies the exact request into the immutable evaluation evidence bundle;
- adds regression coverage preventing workspace-root request references in the recovery job.

The sealed provider artifact, historical evaluator revision, selected alpha, cohort, decision mapping, and information boundary remain unchanged. No provider rerun is authorized.

### DIFF
--- a/.github/workflows/dot-r04-r10-evaluation-recovery-v2.yml
+++ b/.github/workflows/dot-r04-r10-evaluation-recovery-v2.yml
@@ -385,7 +385,17 @@ jobs:
     permissions:
       contents: read
       actions: read
+    env:
+      CONTROL_REQUEST_PATH: control/protocols/execution_requests/dot_r04_r10_evaluation_recovery_v2.json
     steps:
+      - name: Check out exact merged recovery control revision
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.sha }}
+          path: control
+          persist-credentials: false
+          clean: true
+
       - name: Check out exact frozen evaluator revision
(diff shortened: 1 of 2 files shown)


FlorianPfaff:
## Scientific review target

This PR is the continuous-symmetry increment stacked on #464.

The intended paper-level distinction is now explicit:

1. **quotient posterior:** invariant evidence updates physical-orbit masses;
2. **symmetry-complete lift:** prior orbit conditionals are preserved rather than replaced by one gauge representative;
3. **information audit:** any additional within-orbit KL is reported as orbit specificity;
4. **continuous certificate:** a declared `rho`-net and Lipschitz bound cover every unsampled compact-group element;
5. **decision layer:** a finite action may be certified even when a non-invariant physical query remains unresolved.

The controlled SO(2) evidence intentionally uses the tangent-null query `cos(theta)` at `theta=0`: local first-order width is zero, while the complete orbit range is `[-1,1]`. Uniform grids from 4 to 64 samples retain full-range coverage, shrink the explicit `L rho` term, leave an invariant query point-identified, and certify a dominating downstream action.

Promotion to a real continuous-orbit claim still requires rerunning the held Tracking Cloth protocol with registered nested SO(2) grids and a defensible analytic or independently bounded query Lipschitz constant. PR #464 remains the real finite-orbit mechanism evidence; it must not be counted as an independent replication of this controlled extension.

PR #476, opened 2026-09-02
Symmetry-complete beliefs and equivariant decisions without gauge completion

FlorianPfaff:
## Scientific objective

Generalize the finite-orbit/axial-query machinery into a reusable Bayesian physical-twin representation that **never invents a compact-group representative the evidence did not identify**, while avoiding unnecessary abstention when a downstream action is physically coupled to the same unresolved gauge.

The branch now supports four distinct operations:

1. recursively update identifiable quotient information while preserving the unresolved group conditional exactly;
2. answer a fixed query only when it is invariant over the complete orbit;
3. execute a gauge-coupled equivariant action when every compatible group state gives the same or bounded-regret decision; and
4. return the caller-owned exact fallback when the query, action, group scope, or actuator-realization bound is insufficient.

The central separation is:

> the state and fixed world-coordinate query may remain unidentified, while a physically relative action is exactly decision-identifiable.

## 1. Symmetry-complete posterior semantics

The belief is disintegrated as

`p(quotient, group) = p(quotient) p(group | quotient)`.

For invariant evidence, only quotient masses may change. The conditional group law is retained exactly and the KL chain rule is audited:

`KL(joint posterior || joint prior) = KL(quotient posterior || quotient prior) + E_posterior KL(group posterior || group prior)`.

The last term is the **gauge-information expenditure**. It must be zero for an invariant update. Symmetry-breaking evidence uses a separate API and reports positive expenditure instead of disguising it as quotient information.

A joint distribution does not determine `p(group | quotient)` for a zero-mass quotient class. The recursive constructor therefore fails closed unless the caller supplies that continuation conditional explicitly; invariant elimination of a class retains the previously declared conditional rather than silently inserting a uniform or point law.

Point completion is audited rather than normalized away. Selecting one supported representative of a finite orbit adds the quotient-weighted `-log` group probability. A Dirac completion of a continuous conditional density remains singular; finite-quadrature `log(K)` is only a resolution-dependent diagnostic.

## 2. Recursive equivariant Bayesian filtering

An equivariant transition is represented by a quotient transition and a cyclic group-increment kernel:

`p(c_next, g_next | c, g) = T(c_next | c) K(g_next g^{-1} | c, c_next)`.

Prediction is a quotient transition plus circular convolution. Invariant observations repeatedly concentrate `p(c)` without changing `p(g | c)`.

The registered controlled study uses cyclic group sizes 8, 16, 32, and 64. Starting from equal quotient mass and uniform gauge uncertainty, six invariant updates with likelihood ratio `0.8 : 0.2` produce

`p(c=0 | y_1:6) = 4096 / 4097 ~= 0.999756`.

For every group resolution:

- quotient entropy decreases strictly at every update;
- conditional group entropy remains exactly `log(K)`;
- cumulative invariant gauge-information expenditure is zero;
- the same quotient-level action becomes exact zero-regret; and
- an explicit `exp(4 cos(theta))` symmetry-breaking likelihood adds positive group information and reduces group entropy.

Thus this is recursive task identification without recursive gauge hallucination, not only a one-step posterior construction.

## 3. Complete-orbit query certificates

Vector-valued queries retain one shared group draw across all points and coordinates. Independent per-point gauges are not substituted for one common physical latent transform.

For a certified `rho`-net and an `L`-Lipschitz query,

`D_sample <= diameter(q(G)) <= D_sample + 2 L rho`.

The implementation distinguishes `certified-invariant`, `certified-variant`, `undetermined`, and `scope-not-certified`; only certified invariance authorizes a fixed query answer. Exact finite groups are the `rho=0` case.

## 4. Gauge-coupled equivariant decisions

For quotient class `c`, group element `g`, and action template `a`, define

`L_c(g,a) = ell(g x_c, g a)`.

The action need not be world-frame invariant. It is sufficient that every pairwise action-loss difference be invariant:

`L_c(g,a) - L_c(g,b) = L_c(e,a) - L_c(e,b)`.

Then the Bayes-action set and regret vector are independent of the unresolved conditional law `p(g | c)`.

For sampled group nodes and a certified cover, the complete-orbit pairwise bound is

`Delta_bar(a,b) <= sum_c lambda_c [max_s D_cab(g_cs) + K_cab rho_c]`,

with `Reg_bar(a) <= max_b Delta_bar(a,b)`.

The implementation reports sampled and complete-orbit pairwise bounds, decision-equivariance defects, worst-case regret bounds, robustly optimal and epsilon-admissible action sets, deterministic minimax selection, and exact fallback.

## 5. Realized-intervention robustness

The ideal shared-gauge action `g a` is not assumed to be executed perfectly. If Causal4D or another intervention layer supplies

`d_A(u_a(g), g a) <= epsilon_ca`

and the loss is `K_ca`-Lipschitz in action, then the realized pairwise gap receives the explicit margin

`K_ca epsilon_ca + K_cb epsilon_cb`.

The final decomposition is

`realized regret bound = sampled structural gap + continuous-orbit margin + intervention-realization margin`.

The margin is exposed independently rather than hidden inside a generic covariance. Large actuator uncertainty converts exact optimality into a declared bounded-regret result or forces exact fallback.

In the registered sweep with ideal losses `(0, 2, 1)` and unit action-Lipschitz loss:

- realization radii 0.0, 0.2, 0.8, and 1.0 preserve exact optimality;
- radius 1.2 gives a realized regret upper bound of 0.2 and forces fallback at zero tolerance;
- the same radius is admitted as explicitly bounded regret under a separately registered tolerance of 0.25.

## 6. Adversarial strictness result

The state orbit is

`x(theta) = R(theta) [1,0]`

with action templates `R(theta)[1,0]`, `R(theta)[0,1]`, and zero fallback. The state-coordinate orbit has diameter two, so the state is not identified. Under the common physical gauge, losses are constant at `0`, `2`, and `1`; the tracking template therefore has exact zero regret over the complete `SO(2)` orbit.

Two negative controls expose distinct failures:

- a world-frame point completion has worst loss 4 rather than fallback loss 1 and is harmful on two thirds of the orbit;
- independently varying state and action gauges destroys the shared physical cancellation, cannot certify the useful action at zero tolerance, and returns fallback.

## 7. Public real-trajectory evidence and boundary

The merged Tracking Cloth finite-orbit result remains the motivating real-data result: on 15 held recordings and 1,803 controlled geometry cases, a representative-local gate admitted every radial query and about 65.1% of those updates were harmful, while the complete-orbit gate rejected all radial queries, admitted all invariant queries, produced zero harmful accepted radial updates, and returned exact fallback.

The separate continuous calibrated `SO(2)` study retains its negative utility result: support calibration was valid, but rejection exposed a poor fallback. This demonstrates that symmetry validity alone does not make abstention beneficial; fallback competence and direct decision-risk calibration are separate requirements.

The new recursive, equivariant-action, and realized-intervention studies are deterministic mechanism evidence. They do not relabel the existing cloth cohort as fresh confirmation.

## Added in the strengthened version

- `src/prob4d/equivariant_decision.py`;
- `src/prob4d/recursive_symmetry_filter.py`;
- `src/prob4d/realized_equivariant_decision.py`;
- focused adversarial and recursive test suites;
- three content-addressable controlled study protocols and deterministic study scripts;
- `docs/equivariant-decision-identifiability.md`;
- `docs/recursive-and-realized-symmetry-complete-twins.md`;
- permanent Python 3.10/3.12/3.14 workflows with byte-identical replay.

The exact remote head passes the complete Prob4D test suite and all workflows triggered for that head, including the symmetry-complete decision and realized-intervention workflows.

## Claim boundary

This branch does not:

- discover or validate a physical symmetry group;
- infer whole-group invariance from sampled equality;
- validate the state-to-action transform of a robot interface;
- estimate or target-calibrate action-realization radii;
- turn a finite quadrature into exact Haar integration for arbitrary functions;
- validate a learned visual provider;
- establish unseen-object or source-to-target performance;
- identify the complete physical state;
- establish target-domain safety or state of the art.

The decisive empirical promotion is an object- or recording-disjoint public-data study whose provider emits a source-qualified shared state/action transform and whose intervention layer supplies a separately calibrated realization radius. It must compare gauge-coupled actions against MAP completion, independent-gauge uncertainty, invariant-only rejection, covariance inflation, and exact fallback.


FlorianPfaff:
## Cross-repository action-provenance companion

The symmetry-complete result now has a Causal4D companion on branch `research/shared-gauge-intervention-receipt-v1`.

It closes the algebraic interface between an unresolved Prob4D state orbit and a physically transformed command:

- validates one finite multiplication table and separate metric state/action representations;
- requires the action orbit to use the exact same ordered group-element identities as the state orbit;
- independently recomputes `u_cmd(g,a) = R_action(g) a`;
- measures the maximum realized-command discrepancy in the registered action metric;
- rejects an undersized declared radius;
- exports `K_a epsilon_a + K_b epsilon_b` as the pairwise realization-regret margin; and
- binds state evidence, transform instance, action bank, command, realization, loss, fallback, radius scope, and radius provenance into a content-addressed receipt.

The controlled `C4` receipt study reproduces the expected radius `(0.05,0,0)` and pairwise margin

```text
[[0.0, 0.1, 0.1],
 [0.1, 0.0, 0.0],
 [0.1, 0.0, 0.0]]
```

while rejecting a permuted group order, an incorrectly transformed command orbit, and a radius below the observed realization error. A clean clone passes the focused contract tests, byte-identical study replay, and the complete Causal4D test suite; the branch's dedicated GitHub workflow is green.

This creates the intended proof chain:

`Prob4D quotient/group belief -> Causal4D shared-gauge intervention receipt -> Prob4D realized regret bound -> BayesianPhysTwin act or exact fallback`.

The receipt is still a supplied finite-group/realization-envelope certificate. It does not validate the true physical symmetry, continuous-group coverage, unseen-target actuator error, or deployment safety.

FlorianPfaff:
## Full execution chain now implemented

The symmetry-complete branch is now consumed downstream rather than ending at a theorem or local decision object:

1. Prob4D computes the complete-orbit structural pairwise action bound without selecting a gauge representative.
2. Causal4D PR #488 emits a content-addressed portable receipt binding state and action to one ordered group transform and exporting the realized-intervention margin.
3. BayesianPhysTwin PR #883 independently verifies that receipt, binds state/action/loss/fallback identities, adds optional transport margin, and returns an admissible minimax action or exact fallback.

A permanent BayesianPhysTwin workflow checks out all three research branches, records their exact heads, runs the actual Causal4D receipt producer, computes the Prob4D structural bound, and executes the BayesianPhysTwin gate twice with byte-identical output. Action 0 is verified exact-optimal in the valid chain; tampering with the receipt margin produces `invalid-fail-closed` and fallback action 2.

All focused and complete repository suites and the new exact-head workflows are green. The remaining promotion is empirical: an object- or recording-disjoint physical dataset must validate the shared transform and calibrate realization/transport margins before target outcomes are opened.

PR #478, opened 2026-09-02
Add globally optimal posterior rank-distortion frontier

FlorianPfaff:
## Purpose

Extend the exact query-sufficient shared-noise-factor theorem from a single zero-distortion rank to a globally optimal rank--distortion frontier within the same orthogonal `U -> U V` factor-projection family.

## Theorem implemented

For full query posterior covariance `P`, latent response `R = U.T S^-1 C.T`, shared precision Gram matrix `G = U.T S^-1 U`, and `M = I-G`, discarding latent subspace `N` contracts the posterior covariance by

`L_N = R.T N (N.T M N)^-1 N.T R`.

The registered dimensionless distortion is

`D(N) = trace(P^-1 L_N)`.

With `B = R P^-1 R.T`, this is the generalized trace quotient

`trace((N.T M N)^-1 (N.T B N))`.

For discarded dimension `d`, the global minimum over all `d`-dimensional subspaces is therefore the sum of the `d` smallest generalized eigenvalues of `B x = lambda M x`. Choosing nested generalized eigenspaces gives an optimal rank--distortion frontier. Zero distortion recovers the existing exact minimum-rank theorem.

## Repeated-eigenvalue boundary

The optimum distortion value is invariant and unique, but the optimizing factor subspace need not be unique when a retained-rank boundary cuts through a repeated generalized-eigenvalue block. Each frontier point therefore reports:

- `boundary_generalized_eigengap`; and
- `optimal_subspace_unique`.

The tests compare factor-covariance representatives across latent reparameterizations only at strict eigengaps, while comparing optimal distortion at every rank. This avoids claiming a canonical basis where the mathematics supplies only an optimal eigenspace family.

## Audits and controls

Each frontier point independently reports:

- closed-form optimal normalized posterior covariance trace contraction;
- audited contraction from the exact downdate identity;
- maximum normalized covariance contraction;
- expected posterior-normalized mean-shift risk; and
- an analytic mean-shift upper bound based on the shared-precision spectrum.

A deterministic realized-model control shows why this is not merely the existing Euclidean SVD ordering: at retained rank one, the current SVD path has distortion `0.4632161428167537`, whereas the generalized-eigen optimum is `0.3703045854595735` (about 25% lower). Random same-rank subspaces are also checked against the analytic optimum.

## Validation

At head `baaebd29d576c77be77b97b75383a3a9dd89eaef`:

- focused formatting, Ruff, mypy, and all seven dense theorem/control tests pass;
- the complete repository suite passes on Python 3.10, 3.11, 3.12, 3.13, and 3.14;
- the declared Python/NumPy runtime-floor suite passes;
- wheel and source-distribution build/install/CLI smoke tests pass;
- authoritative documentation/API compatibility and provider-contract lanes pass; and
- dependency audit plus CodeQL for Python and GitHub Actions pass.

## Scientific boundary

The result is globally optimal only for the registered normalized posterior-covariance trace-contraction objective, within one frozen supplied-factor `U -> U V` family. It does not preserve observation likelihood, claim full posterior KL optimality, establish recursive exactness, or establish end-to-end task/control optimality. Aggressive ranks still require a valid-posterior audit; the complete factor remains the terminal zero-distortion point.

This branch is independent of the separately retained recursive-closure experiment in PR #475, whose real-cloth structural diagnostic showed that strict infinite-horizon exact closure becomes full-state. Real-data rank--distortion evaluation is intentionally being developed in a stacked PR rather than mixed into this theorem/kernel change.


FlorianPfaff:
Stacked public real-data evidence is now available in PR #487.

On 80 admitted cloth recordings (20,843 held-out windows; ten recording-disjoint fitted models), the generalized-eigen frontier strictly reduced the registered posterior trace-contraction objective versus the old response-SVD ordering in every rank-1 and rank-2 fold: mean per-fold reductions 8.23% and 5.70%, respectively. Rank 3 retained exact posterior parity with a 17.01x shared-factor payload reduction.

The important negative result is equally consistent: every rank-0/1/2 reduced query posterior was indefinite, and lower registered contraction did not imply lower held-out RMSE. This motivates the next theorem extension: spectral-minimax optimality and an exact rank-feasibility certificate for Bayesian validity, rather than treating an invalid chosen projection as merely an optimizer defect.

